Issue #12, opened 2019-12-04
null value in parameter


martinxw:
string expression = "if( [value] == '', 'majeur', 'mineur')";
var exp = new NCalc.Expression(expression);
exp.Parameters["value"] = null;
object o = exp.Evaluate();

invoke error inside function CompareUsingMostPreciseType.
Body of this function is:
`public int CompareUsingMostPreciseType(object a, object b)`
`{`
`            Type mpt = GetMostPreciseType(a.GetType(), b.GetType());`
`            return Comparer.Default.Compare(Convert.ChangeType(a, mpt), Convert.ChangeType(b, mpt));`
`}`

I have a little correction:
`public int CompareUsingMostPreciseType(object a, object b)`
`{`
`           Type mpt;`
`            if (a == null)`
`            {`
`                if (b == null)`
`                    return 0;`
`                mpt = GetMostPreciseType(null, b.GetType());`
`            }`
`            else`
`            {`
`                mpt = GetMostPreciseType(a.GetType(), b?.GetType());`
`            }`
`            return Comparer.Default.Compare(Convert.ChangeType(a, mpt), Convert.ChangeType(b, mpt));`
`}`

and now it work also with null

yallie:
Hi @martinxw, 

thanks for your suggestion!
Would you mind creating a pull request?

martinxw:
done

PR #13, opened 2019-12-05
Update EvaluationVisitor.cs

martinxw:
allow null value in parameter described here: https://github.com/ncalc/ncalc/issues/12


yallie:
Great! Thank you :)

PR #14, opened 2020-01-21
Add natural logarithm "Ln" to supported syntax

Andree2:


yallie:
Thanks @Andree2!

Would you mind creating the same PR for [ncalc-async](https://github.com/ncalc/ncalc-async) project?

Andree2:
@yallie Yeah sure, done. I was considering switching to ncalc-async anyway, but as that would take more than an hour to adapt all usages, that had to be postponed.

yallie:
Great, thanks @Andree2! I think ncalc-async is the future, ncalc is kept for backwards compatibility with legacy projects :)

PR #19, opened 2020-11-04
Fixes #18: Handle long integers as long, not single-precision floats

### LINKED ISSUE #18: Long integers are treated as single-precision floats
Reported here: https://github.com/ncalc/ncalc-async/issues/4

petli:
This is a port of https://github.com/ncalc/ncalc-async/pull/5. This may require a new major version, since it will change how expressions with large numbers are being evaluated.


yallie:
Great! Thank you Peter! 👍

PR #28, opened 2021-08-13
Added Expression.GetParameters() method

rodion-m:
This method I used in NCalc 1.*. It's useful to have this method from C++/CLI, because it doesn't support lambda-expressions - so, `EvaluateParameter` is hard to use in C++/CLI.


yallie:
Great! Thank you. 
Do you have a unit test demonstrating the usage?

Does it work with expressions containing custom functions? Something like this:
https://github.com/ncalc/ncalc/blob/master/Evaluant.Calculator.Tests/Fixtures.cs#L125

rodion-m:
> Great! Thank you.
> Do you have a unit test demonstrating the usage?

Sure, let me create a test file for this. `Tests/GetParameters.cs` is ok?

yallie:
Hi @rodion-m, it this PR ready to merge?

>Sure, let me create a test file for this. Tests/GetParameters.cs is ok?

What about the test class?
Have you tried your solution with expressions containing custom functions?
I believe your solution may capture custom function names along with parameters.

yallie:
Superseded by pull request #104

PR #31, opened 2021-09-10
Fixes #16: Fix obvious copy-paste error

### LINKED ISSUE #16: Abs(x)/Pow(y,2) crashes
The error raised is: Operator '/' cannot be applied to operands of type 'decimal' and 'double'. It appears that Abs() is returning a Decimal even though x was a Double.

murksa:


yallie:
Thanks! 👍

Issue #37, opened 2021-12-10
Case insensitive logical operators 


huantianad:
Hey, just dropping in here to ask if it's possible for the `and`, `or`, and `not` operators to be case insensitive, as my current setup already uses `And` and `Or` instead of their lowercase versions.

Ideally custom operators would be nice; however, I understand that that may be unreasonable.

petli:
It is a bit fiddly to support that, since the ANTLR3 grammar that NCalc uses is case-sensitive. There are ways to hack around that, though: https://stackoverflow.com/a/22160240

yallie:
Implemented by @petli in v3.0.

Issue #36, opened 2021-12-09
Support ** as POW alternative.


DomRe:
Since ^ is reserved for bitwise, would be good to get support for ** as a unary operation for POW calculations.

DomRe:
Can be closes. Was done by petli in ncalc async.

Issue #35, opened 2021-11-21
How case sensitivity works?


samanthwal:
example 1
var calculation = "( [User] == 'ABC') **OR** ([User] == 'XYZ' ) ";

var nCalcExpression = new Expression(calculation, **EvaluateOptions.IgnoreCase)**;

nCalcExpression.Parameters["User"] = "XYZ";

if (nCalcExpression.HasErrors())
{
	**throw new FormatException($"Could not evaluate expression: '{calculation}' due to {nCalcExpression.Error}.");**

//give error here as it does not recognize "OR" but it does "or'
}

Console.WriteLine(nCalcExpression.Evaluate());

example 2
var calculation = "( [User] == 'ABC') or  ([User] == 'XYZ' ) ";

var nCalcExpression = new Expression(calculation, **EvaluateOptions.IgnoreCase)**;

nCalcExpression.Parameters["User"] = **"xyz";**

if (nCalcExpression.HasErrors())
{
	throw new FormatException($"Could not evaluate expression: '{calculation}' due to {nCalcExpression.Error}.");
}

Console.WriteLine(nCalcExpression.Evaluate());

It evaluates to "FALSE" because user value supplied is "xyz" not "XYZ"

then my question is what is the use of EvaluateOptions.IgnoreCase option?

petli:
`EvaluateOptions.IgnoreCase` only affect functions and parameter lookup by making them case insensitve, it has no effect on strings.

If case-insensitive string comparison is desired, a new option property should be added for that so existing behaviour isn't affected.

petli:
This is now fixed in 3.0, you can use any case for keywords like `Or` or `tRuE` etc.

Issue #21, opened 2020-12-07
Support for numbers with trailing dot


alexkaratzas:
Hi,

some languages like JavaScript and Python support numbers like 1. which get evaluated to a float with the value 1. This currently is not supported by your grammar, could you provide support for it?

Kind regards,
Alex

petli:
This is fixed in version 3.0.0.

PR #38, opened 2021-12-17
Port new syntax features from ncalc-async 3.0

petli:
Implements the following issues:

#11 
#21 
#35
#36
#37


yallie:
@petli, thanks for porting the new features to the sync project!
That's awesome. You even refactored the grammar so that it doesn't need manual tweaks anymore :)

One minor thing.
I'm not sure about the priority of the negation operator compared to the exponentiation operator.
C# don't have power operator, so we can't take it as a reference implementation.

Visual Basic power operator seems to have higher priority than unary negation:
https://docs.microsoft.com/en-us/dotnet/visual-basic/language-reference/operators/operator-precedence

```vb
Imports System

Class Program
	Shared Sub Main()
		Console.WriteLine(-2 ^ 2) ` prints -4, which means it's evaluated as -(2 ^ 2) 
	End Sub
End Class
```
Javascript unary negation has higher priority than power operator:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Operator_Precedence

But Chrome, Edge and Mozilla Firefox treat the following expression as a syntax error:

```js
>>  -1 ** 2
/// Firefox: SyntaxError: unparenthesized unary expression can't appear on the left-hand side of '**'
/// Chrome: Uncaught SyntaxError: Unary operator used immediately before exponentiation expression. Parenthesis must be used to disambiguate operator precedence
```

What do you think? :)

yallie (commented):
Looks awesome 👍

petli:
@yallie Since C# doesn't have the `**` operator, I based it on the expression priority in Python (the next language I know where it exists). 

I didn't know about the Javascript exception, but given that all major browsers warn the user about the odd precedence and requires parentheses I think we are safe sticking to the Python/VB priority :)

PR #47, opened 2022-03-30
Make Evaluate method virtual

Shifty15:
We need to override Evaluate method because we need to do some staff before and after evaluate method.
Example:
public override object Evaluate()
        {
            errors.Clear();
            var result = base.Evaluate();
            //return null if there was Evaluation error
            return errors.IsNullOrEmpty() ? result : null;
        }

Vladimir Jaron

### DIFF
--- a/src/NCalc/Expression.cs
+++ b/src/NCalc/Expression.cs
@@ -206,7 +206,7 @@ namespace NCalc
         protected Dictionary<string, IEnumerator> ParameterEnumerators;
         protected Dictionary<string, object> ParametersBackup;
 
-        public object Evaluate()
+        public virtual object Evaluate()
         {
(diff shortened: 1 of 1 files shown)


yallie:
Hi Jaron, 

what's the purpose of subclassing the `Expression`?
`Expression.Compile("2+3")` won't return your derived class anyway, and Evaluate will not be called.
Did you consider wrapping around NCalc with your own façade?

NCalc integrated within the host app usually provides custom functions, parameter hooks, etc.
Custom wrapper class is a good place to set up such infrastructure specific to your app.

Shifty15:
Hi,
 Thank you for your fast reply. We will use wrapper of Expression.

Thank you.
Best
Vladimir Jaron

PR #52, opened 2022-05-30
Allow custom culture info fixes

rholek:
Fixes for https://github.com/ncalc/ncalc/pull/46. Culture info was not passed to all methods correctly.

### DIFF
--- a/src/NCalc/Domain/EvaluationVisitor.cs
+++ b/src/NCalc/Domain/EvaluationVisitor.cs
@@ -72,7 +72,7 @@ namespace NCalc.Domain
                 mpt = GetMostPreciseType(a.GetType(), b?.GetType());
             }
             
-            return Comparer.Default.Compare(Convert.ChangeType(a, mpt, _cultureInfo), Convert.ChangeType(b, mpt));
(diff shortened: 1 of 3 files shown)


yallie:
Thanks for the PR @rholek!

Could you please add a couple of unit tests that demonstrate the difference?
I.e. a couple of expressions that doesn't calculate when cultureInfo is not specified but work fine if it's specified.
Something like this —

```csharp
[Test]
public void TestCultureInfo()
{
    // a culture-specific expression example
    var expr = "...";

    // throws if CultureInfo is not specified
    Assert.Throws<EvaluationExpression>(() => Expression.Evaluate(expr));

    // throws if CultureInfo is not the expected one
    Assert.Throws...

    // doesn't throw if CultureInfo is specified
    var result = ...
    Assert.AreEqual(..., result);
}
```

rholek:
I added the test

yallie:
Thank you @rholek! 👍

rholek:
@yallie please publish this fix to nuget

yallie:
Thanks @rholek, here is it —
https://www.nuget.org/packages/NCalcSync/3.4.0

PR #55, opened 2022-06-10
Using OrdinalIgnoreCase instead of ToLower() checks

Bykiev:
Using OrdinalIgnoreCase instead of ToLower() checks. It's the preferred way for comparing strings, also it could improve performance in some cases.


yallie:
Thanks Andrey!
Sorry for my late response, completely missed the notification about this PR.

PR #59, opened 2022-11-22
Add arithmetic support for multiple operators and types

baughj:
This PR attempts to fix a number of inconsistencies I found while using the library:

1. Most operands did not support `byte`. This has been fixed. 
2. Implemented unit tests for `Add`, `Subtract`, `Multiply`, and `Modulo`.
3. There were inconsistencies between supported types and operands. As much as possible, this has been fixed.
4. `Soustract` was changed to `Subtract` as the rest of the functions are in English. I am not sure if this is part of an exposed API; if it is I will revert it. As far as I can tell `Soustract` was/is only used by `EvaluationVisitor`.
5. Various misspellings were corrected and if the type was wrong in the error message reported upwards that was also fixed.
6. Default handling was put in switches to make it more ergonomic for users (eg a usable error). Right now if you evaluate an expression with variables that resolves to a type that is not supported in `Numbers` it is very difficult to debug as a `NullExceptionError` is thrown from deep in the library.

I am not sure I understand division enough yet to make it match the other operands. It is not like the others in that it modifies its inputs before evaluation, and it seems some of the tests rely on the fact that it only converts the left hand side of something to a `double` if it is not decimal / single / double (eg `IsReal()` usage in `EvaluationVisitor`). 

Coercing both lhs / rhs to `Double` based on what `IsReal` returns broke a number of other tests so I am leaving it as is and not implementing a test for division in this PR. That said, this means there are some unusual consequences of how variable division works currently:

```csharp
var expression = "X / Y";
Expression e = new(expression);
e.Parameters["X"] = true;
e.Parameters["Y"] = 1;
var ret = e.Evaluate(); // Evaluates to 1 (eg 1/1 = 1)
```

vs

```csharp
var expression = "X / Y";
Expression e = new(expression);
e.Parameters["X"] = 1;
e.Parameters["Y"] = true;
var ret = e.Evaluate(); // Throws InvalidOperationException - Operator '/' can't be applied to operands of types 'double' and 'bool'
```

Thanks a lot for your work on this library! I hope this is a useful PR.

### DIFF
--- a/src/NCalc/Domain/EvaluationVisitor.cs
+++ b/src/NCalc/Domain/EvaluationVisitor.cs
@@ -166,7 +166,7 @@ namespace NCalc.Domain
                     break;
 
                 case BinaryExpressionType.Minus:
-                    Result = Numbers.Soustract(left(), right(), _cultureInfo);
+                    Result = Numbers.Subtract(left(), right(), _cultureInfo);
                     break;
 
                 case BinaryExpressionType.Modulo:
@@ -232,7 +232,7 @@ namespace NCalc.Domain
                     break;
 
                 case UnaryExpressionType.Negate:
-                    Result = Numbers.Soustract(0, Result, _cultureInfo);
+                    Result = Numbers.Subtract(0, Result, _cultureInfo);
                     break;
 
                 case UnaryExpressionType.BitwiseNot:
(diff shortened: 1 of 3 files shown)


baughj:
Fixed the modulo test, was using wrong operand 🤦 
Also worked around the handling of the add test, since add also tries to convert its operands first.
Both are passing now.

yallie:
Awesome! Thanks Justin!

Issue #60, opened 2022-11-26
Performance / Cache


ThomasHambach:
Hi!

First of all, thanks for such a well-maintained package. I have been a happy user for many years.

However, I do have a question. We're having about 20-30 expressions parsed on a request to an API endpoint. The first time (no cache) would take around 1-3 seconds and afterward it would take only 80ms as they are cached. Unfortunately, running this in the cloud means each instance of the server has to create its own cache. Would it be feasible for me to alter the caching layer and place it in Redis?

Besides that, I am also thinking of creative ways to "warm up" the cache. Such as creating the cache entry as soon as the expression is written. But I am unsure whether they may be pitfalls or if this is just me being too optimistic.

Thanks a lot!

ThomasHambach:
Actually, I have (ahem) solved it. The only thing you would need to do is add a parameterless constructor to the `ValueExpression` class. After that, you can use Newtonsoft to serialize the `LogicalExpression`.

```
public class ValueExpression : LogicalExpression
{

    public ValueExpression() {}
```

You can serialize it as shown below if anyone is interested. No guarantees on whether this will work with more complex expressions, I will report back on that at a later date.

```
var compiled = Expression.Compile(expression, true);
var json = JsonConvert.SerializeObject(compiled, new JsonSerializerSettings
{
    TypeNameHandling = TypeNameHandling.All // We need this to allow serializing abstract classes
});
```

yallie:
Hi @ThomasHambach, sorry for my late response.

I've never tried to serialize the compiled expressions, but it sounds like a reasonable approach.
I don't see anything wrong with it as long as your servers are identical (same NCalc version, same server logic, etc).
Please let me know if you find any peculiar behavior with shared cached expressions.

Cheers!

ThomasHambach:
Don't worry about it, you're maintaining open-source software and I am happy that you're actively checking. It would however be great if we can add the parameterless construction to this package so I do not have to keep updating my fork. Happy to make a PR for it.

One more thing I found in comparison to `ncalc2` is that the caching doesn't work in a multithreaded context. Refer to commit https://github.com/sklose/NCalc2/issues/22 
I am unsure if this is a real issue at this time. I see the change is not applied in this repo.

yallie:
>It would however be great if we can add the parameterless construction to this package so I do not have to keep updating my fork. Happy to make a PR for it.

Great! Will appreciate it.

>I am unsure if this is a real issue at this time. I see the change is not applied in this repo.

[...]

PR #61, opened 2022-11-29
Add parameterless constructor to ValueExpression

ThomasHambach:
Add a parameterless constructor to ValueExpression that will allow serialization of the class. Will create a separate PR for the cache changes.

With this change, you can use Newtonsoft to serialize the `LogicalExpression`.

```
var compiled = Expression.Compile(expression, true);
var json = JsonConvert.SerializeObject(compiled, new JsonSerializerSettings
{
    TypeNameHandling = TypeNameHandling.All // We need this to allow serializing abstract classes
});
```

Also, see https://github.com/ncalc/ncalc/issues/60

### DIFF
--- a/src/NCalc/Domain/Value.cs
+++ b/src/NCalc/Domain/Value.cs
@@ -4,6 +4,8 @@ namespace NCalc.Domain
 {
     public class ValueExpression : LogicalExpression
     {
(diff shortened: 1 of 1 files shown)


yallie:
Thanks!

PR #63, opened 2023-01-19
Add Min and Max to unit tests

ThomasHambach:
While porting the library to Typescript, I noticed `Min` and `Max` are not covered in the unit tests. This is a small change that adds these tests.


yallie:
Thanks!

PR #69, opened 2023-05-17
Abs convert to double

PPRhuM:
Convert "Abs" function parameter to double to stay consistent with other functions (they're all using double parameters).


PPRhuM:
Sorry, I forgot to change unit tests

PR #78, opened 2023-07-16
Fix issue with CommonTypes

ThomasHambach:
As mentioned in my issue, this PR fixes the order of the CommonTypes to consider Double before an Int64. This may have some cost on the performance and we may want to revisit how the type detection works in the future.

I have also added an example to the README of how to use this with a distributed cache (eg, Redis).

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,6 +85,35 @@ e.EvaluateParameter += delegate(string name, ParameterArgs args)
 Debug.Assert(117.07 == e.Evaluate());
 ```
 
+**Caching in a distributed cache**
+
+This example uses [Newtonsoft.Json](https://www.nuget.org/packages/Newtonsoft.Json/).
+
+Serializing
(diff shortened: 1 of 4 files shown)


yallie:
Great! Thank you Thomas! 👍

ThomasHambach:
I don't think I have any rights to publish a patch version. Would it be possible to release? Or, alternatively, point me how to do so. :)

yallie:
@ThomasHambach, 

I've changed your role to the Owner.
Now you should be able to push commits and tags.
Please try following this procedure to publish `v3.8.0` release:

https://github.com/ncalc/ncalc/wiki/Publishing-a-new-release

Adding a release should automatically push the new version of the Nuget package.
Thanks!

yallie:
https://www.nuget.org/packages/NCalcSync/3.8.0

PR #90, opened 2023-12-11
Added EvaluationVisitor at ctor

### LINKED ISSUE #89: Expression should take have a constructor that takes in an EvaluationVisitor
https://github.com/ncalc/ncalc/blob/9f28e7b31ddfe42f8cd74128acb0bddc5511046f/src/NCalc/Expression.cs#L243

If you have a custom EvaluationVisitor class,  you cannot use Expression.Evaluate because the visitor is local to the method. That means you need to also virtually re-implement the entire Expression class

gumbarros:
This closes #89

### DIFF
--- a/src/NCalc/Expression.cs
+++ b/src/NCalc/Expression.cs
@@ -24,6 +24,8 @@ public class Expression
     /// Get or set the culture info.
     /// </summary>
     protected CultureInfo CultureInfo { get; set; }
+    
+    protected EvaluationVisitor EvaluationVisitor { get; set; }
 
     public Expression(string expression) : this(expression, EvaluateOptions.None, CultureInfo.CurrentCulture)
     {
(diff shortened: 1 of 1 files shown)


thetreatment:
I think for EvaluationVisitor to be sub-classable, it also needs the OnEvaluateParameters and OnEvaluateFunction to be changed to protected (from private), otherwise there's no way to invoke the handlers from a sub-class (since you can't raise the parent event).

For example:
```
class CustomVisitor: EvaluationVisitor {

public override void Visit(Identifier parameter) {
if (something){
     // custom logic
    var args = new ParameterArgs();
     OnEvaluateParameters(parameter.Name, args);  <-----not accessible
    //....
}
else
     base.Visit(parameter);
}
```

thetreatment:
While attempting to subclass EvaluationVisitor, the other issue is Result is a private set.

yallie:
Hi @gumbarros and @thetreatment,

sorry for my late reply!
Just got back from my vacation. 

Thanks for the PR, merging.

yallie:
@gumbarros oh, my bad! Sorry.

An incomplete pull request is better be marked as a draft, or I'll just merge it right away 😁

@thetreatment,

>I think for EvaluationVisitor to be sub-classable, it also needs the OnEvaluateParameters 
>and OnEvaluateFunction to be changed to protected (from private)

Looks reasonable! Could you please make a PR with this change?

thetreatment:
Done: https://github.com/ncalc/ncalc/pull/92

PR #91, opened 2023-12-14
Add support for ifs(cond, val, default)

baughj:
This PR adds support for  a new function`ifs()`, similar to `if()` but allows an arbitrary collection of `(condition, value)` pairs along with a default value if none of the conditions evaluate to true.

Examples:

`ifs([divider] != 0, [divided] / [divider], -1)` -> if `divider` != 0, return the value of `[divided] / [divider]`, otherwise return -1

I added unit tests for this as well, let me know if you think they need to be more comprehensive.

### DIFF
--- a/src/NCalc/Domain/EvaluationVisitor.cs
+++ b/src/NCalc/Domain/EvaluationVisitor.cs
@@ -2,6 +2,7 @@
 using System.Collections;
 using System.Collections.Generic;
 using System.Globalization;
+using System.Linq;
 
 namespace NCalc.Domain;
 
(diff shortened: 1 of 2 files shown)


baughj:
@yallie any thoughts on above?

yallie:
Hi @baughj, 

sorry for my late reply!
Just got back from my vacation. 
Thanks for the PR, merging.

baughj:
Awesome, thank you!

baughj:
@yallie one thing - how do i go about updating documentation?

yallie:
@baughj I'm not sure about wiki access rights here on github...

Please have a look at this wiki page: https://github.com/ncalc/ncalc/wiki/Functions
Do you see this `Edit` button in the upper-right corner?

image

baughj:
Yep! It works. Sorry for the delay, I updated the documentation.

PR #92, opened 2023-12-21
Make EvaluationVisitor sub-class friendly

thetreatment:
OnEvaluateParameters and OnEvaluateFunction to be changed to protected (from private), otherwise there's no way to invoke the handlers from a sub-class (since you can't raise the parent event).

Also adding public access to the original string expression and moving to protected set. This is a convenience over having to compile the expression and then call ToString() on Expression just for a string representation.


yallie:
Thank you! 👌

yallie:
@thetreatment, please consider joining ncalc organization :)

PR #97, opened 2024-01-09
Allow Override EncapsulateNoValue in SerializationVisitor

### LINKED ISSUE #96: FR: Allow Override EncapsulateNoValue in SerializationVisitor
I would like to subclass SerializationVisitor and make some minor changes to EncapsulateNoValue, but since it is not virtual, I cannot override it. As a workaround I have to copy the entire class.

As a feature, can you mark this function as virtual?

https://github.com/ncalc/ncalc/blob/bb4f9cb44422625ed332725eed8c34759bd26f90/src/NCalc/Domain/SerializationVisitor.cs#L187C9-L187C9

gumbarros:
Also removed unused private static method.

This closes #96

### DIFF
--- a/src/NCalc/Domain/LogicalExpression.cs
+++ b/src/NCalc/Domain/LogicalExpression.cs
@@ -1,47 +1,7 @@
-using System;
-using System.Text;
-
 namespace NCalc.Domain;
 
 public abstract class LogicalExpression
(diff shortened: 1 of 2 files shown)


yallie:
Thanks! 👍

PR #98, opened 2024-01-10
ignore build result files

hhuang2013:
.gitignore changed to ignore build result files in folder 'obj' and 'bin'


yallie:
Thanks!

PR #104, opened 2024-02-22
Add GetParametersNames method to Expression.cs

### LINKED ISSUE #103: How to get a list of all variables (or their names) used in an expression?
Alternatively: is it possible to make it so that if the value of some parameter was not defined, some default value would be used instead of throwing an exception?
Context: I'm using NCalc expressions as part of an in-game programming language, and I need to be sure that even if the user didn't set some variables before evaluating the expression, they'll get a result and not an error.

gumbarros:
This closes #103

### DIFF
--- a/src/NCalc/Domain/ParameterExtractionVisitor.cs
+++ b/src/NCalc/Domain/ParameterExtractionVisitor.cs
@@ -1,0 +1,37 @@
+﻿using System.Collections.Generic;
+
+namespace NCalc.Domain;
+
+internal class ParameterExtractionVisitor : LogicalExpressionVisitor
+{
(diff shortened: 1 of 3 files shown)


gumbarros:
@yallie you can also close PR #28 that does basically the same thing

yallie:
Hi @gumbarros, thanks for the PR!

But what if an expression has custom functions?
Doesn't it list custom functions as parameters?

gumbarros:
Hmmm, I will cover this in a unit test to be sure

gumbarros:
@yallie everything looks fine, tried both custom functions and native functions

yallie:
Thanks!

Issue #109, opened 2024-03-04
Null handling changed from version 2 to 3


Coder3333:
I have this code in my application, and it worked properly with every version of NCalc prior to 3.0.  Now, this same code throws an exception when evaluating the expression.

```c#
            var expression = "{null} = 'WrongValue'";
            var eap = new NCalc.Expression(expression
    // need to indicate NoCache in order to avoid threading issues
    , NCalc.EvaluateOptions.NoCache);

            eap.EvaluateParameter += delegate (string name, NCalc.ParameterArgs args)
            {
                if (name == "null")
                {
                    // if 'null' appears in an expression, it is considered a parameter, so must replace it with
                    // a string.  This will be an issue if comparing "null = 'null'", as it would evaluate to true, instead
                    // of false.  will need to investigate if there is a better way to handle null
                    args.Result = "null";
                }
            };
            var evaluated = eap.Evaluate();
```

Exception:
```
  Message: 
NCalc.EvaluationException : token recognition error at: '{':1:0
token recognition error at: '}':1:5
---- NCalc.EvaluationException : token recognition error at: '{':1:0
token recognition error at: '}':1:5

  Stack Trace: 
Expression.Evaluate()
ExpressionTests.DemonstrateNullException() line 92
----- Inner Stack Trace -----
Expression.Compile(String expression, Boolean nocache)
Expression.HasErrors()

``

yallie:
@Coder3333, you probably used another NCalc fork before. This repository publishes [NCalcSync](https://www.nuget.org/packages/NCalcSync) package.

@gumbarros By the way, NCalc should support square brackets surrounding the parameters.
Comes in handy for parameter names containing spaces and/or unicode characters.

Coder3333:
> @Coder3333, you probably used another NCalc fork before. This repository publishes [NCalcSync](https://www.nuget.org/packages/NCalcSync) package.
> 
> @gumbarros By the way, NCalc should support square brackets surrounding the parameters. Comes in handy for parameter names containing spaces and/or unicode characters.

@yallie , aha.  I have been using https://www.nuget.org/packages/CoreCLR-NCalc/.  Didn't realize I had the wrong repo.  Any idea if I should be able to switch to this package without too much work?

yallie:
[...]

PR #112, opened 2024-03-08
Add Curly Braces support

gumbarros:
Related to #109


gumbarros:
Don't know if this will overcomplicate things, feel free to re-open if you think this is useful.

PR #116, opened 2024-03-09
Use correct CLR types at GetMostPreciseType

### LINKED ISSUE #102: CompareUsingMostPreciseType fails on Int32
If a==null and b.GetType() is Int32, the comparison fails because it is not recognized as a basic type. Only Int64 is included in that set.

gumbarros:
This closes #102


src/NCalc/Domain/EvaluationVisitor.cs
```diff
@@ -699,11 +715,8 @@ public override void Visit(Identifier parameter)
         if (Parameters.ContainsKey(parameter.Name))
```
Bykiev, line 715: you can replace `ContainsKey` with `TryGetValue` and then use the retrieved value on line 718 and 729

yallie: Good point! Thanks

Beckdotan:
can you please return CompareUsingMostPreciseType to be public? I'm using it since i wrote a new visit function in this code:
https://dotnetfiddle.net/5iSl8A
Since I wanted to create my own logic for null values.

gumbarros:
@Beckdotan sorry for the unintended change, I published 3.13.1 with the fix

PR #118, opened 2024-03-09
Allow using decimal as default floating point type

### LINKED ISSUE #117: Allow using decimals as default for floating points
Hey,

for a use-case I'm working on it would be very helpful to allow specifying the default type for floating point numbers as `decimal` instead of `double`. When specifying this setting, floating point rounding errors can be avoided in many cases, at the cost of performance. Since decimals are already supported the changes within the code base should not be too large from my first investigation. 

Is this a feature that would be of interest for merging? I'll happily prepare a PR for this.

lucaschimweg:
This fixes #117 

- Replace parameter nocache with passing EvaluationOption in Expression.Compile, but keep backwards compatibility
- Add EvaluationOption.HasOption extension method for readability


lucaschimweg:
Note: The large diffs in the Parser & Lexer are only whitespace changes. The only actual change is the ANTLR version (used latest version) and the two changes reflected in the grammar. I had to change the language from `CSharp3` to `CSharp` since ANTLR does not seem to support `CSharp3` anymore

gumbarros:
Looks very LGTM, will probably conflict with my other PR about Curly Braces that also updates ANTLR.

Also nice extension method.

Btw, can a maintener approve the PRs? looking forward to use the changes at Monday @yallie @ThomasHambach @petli ?

lucaschimweg:
Thanks! 
If we merge your other PR first I can happily merge back master and rebuild the ANTLR.

yallie:
Looks great! Merging. 

Thank you!

yallie:
Release: [v3.12.0](https://github.com/ncalc/ncalc/releases/tag/v3.12.0)
Package: [v3.12.0](https://www.nuget.org/packages/NCalcSync/3.12.0)

PR #123, opened 2024-04-21
Fix boolean comparison

### LINKED ISSUE #122: Comparing String and Bool - changed behavior - wanted ?
Hello,

between Version 3.11 and 3.12 the behavior comparing for example bool and string change.
I only want to know if this behavior is now the default (and I must find a workaround), or if this could be made a little bit more user friendly ?

If You compare a string with a bool ... GetMostPreciseType says - compare it by string.
So the bool is converted to a String using Convert.ChangeType.
This produce  true -> "True".

Now the comparing of variables of type string is case-sensitive.

The Expression:

"foo = true"
(with foo="true")

produce false , in 3.11 this was true.
(internal it compares "true"=="True")

I know - I can enable case insensitive compare for all String compares.
But it is a idea to handle boolean different.... ? 
For example on Compare String with Bool - to convert the String into bool.
(TryParse ?) this would handle True,true,1 to bool ?

Is the current behavior change - a unnoticed regression bug - or its a feature to force user for more explicit expressions with different types ? ;)

gumbarros:
This closes #122

### DIFF
--- a/src/NCalc/Domain/EvaluationVisitor.cs
+++ b/src/NCalc/Domain/EvaluationVisitor.cs
@@ -51,6 +51,7 @@ public class EvaluationVisitor(EvaluateOptions options, CultureInfo cultureInfo)
         typeof(byte),
         typeof(sbyte),
         typeof(char),
+        typeof(bool),
         typeof(string),
         typeof(object)
     ];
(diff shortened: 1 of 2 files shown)


yallie:
Great! Thanks a lot for fixing that! 👍

PR #126, opened 2024-04-22
Performance: Use pattern matching instead of `TypeCode`

gumbarros:
Here is a benchmark :)

```cs
using BenchmarkDotNet.Attributes;
using BenchmarkDotNet.Mathematics;
using BenchmarkDotNet.Running;
using ValueType = Benchmark.ValueType;

[RankColumn(NumeralSystem.Arabic)]
[MemoryDiagnoser]
public class ValueExpressionBenchmark
{
    private object value = 42.5; 

    [Benchmark]
    public void TypeCodeChecking()
    {
        switch (System.Type.GetTypeCode(value.GetType()))
        {
            case TypeCode.Boolean:
                Type = ValueType.Boolean;
                break;
            case TypeCode.DateTime:
                Type = ValueType.DateTime;
                break;
            case TypeCode.Decimal:
            case TypeCode.Double:
            case TypeCode.Single:
                Type = ValueType.Float;
                break;
            case TypeCode.Byte:
            case TypeCode.SByte:
            case TypeCode.Int16:
            case TypeCode.Int32:
            case TypeCode.Int64:
            case TypeCode.UInt16:
            case TypeCode.UInt32:
            case TypeCode.UInt64:
                Type = ValueType.Integer;
                break;
            case TypeCode.String:
                Type = ValueType.String;
                break;
            default:
                throw new InvalidOperationException("This value could not be handled: " + value);
        }
    }

    [Benchmark]
    public void PatternMatching()
    {
        Type = value switch
        {
            bool => ValueType.Boolean,
            DateTime => ValueType.DateTime,
            decimal or double or float => ValueType.Float,
            byte or sbyte or short or int or long or ushort or uint or ulong => ValueType.Integer,
            string => ValueType.String,
            _ => throw new InvalidOperationException("This value could not be handled: " + value)
        };
    }

    public ValueType Type { get; private set; }

    public void RunBenchmark()
    {
        var summary = BenchmarkRunner.Run<ValueExpressionBenchmark>();
    }
}
```
image

### DIFF
--- a/src/NCalc/Domain/EvaluationVisitor.cs
+++ b/src/NCalc/Domain/EvaluationVisitor.cs
@@ -107,12 +107,7 @@ public class EvaluationVisitor(EvaluateOptions options, CultureInfo cultureInfo)
         }
     }
 
-    private static bool IsReal(object value)
-    {
-        var typeCode = Type.GetTypeCode(value.GetType());
-
-        return typeCode is TypeCode.Decimal or TypeCode.Double or TypeCode.Single;
-    }
+    private static bool IsReal(object value) => value is decimal or double or float;
 
     public override void Visit(BinaryExpression expression)
     {
--- a/src/NCalc/Domain/Value.cs
+++ b/src/NCalc/Domain/Value.cs
@@ -14,40 +14,15 @@ public class ValueExpression : LogicalExpression
 
     public ValueExpression(object value)
     {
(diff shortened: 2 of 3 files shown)


yallie:
Wow, perf difference is amazing... I thought it would be 10-30% at most.

yallie:
Thanks! 🔥

PR #127, opened 2024-04-22
Test for inner exception type instead of exception message

### LINKED ISSUE #120: Unit tests fails with locale other than English
This unit test fails on systems where the locale is other than English:
https://github.com/ncalc/ncalc/blob/08994e2c366d4f2621aeea461c0e12ac1448dbf4/test/NCalc.Tests/Fixtures.cs#L712
The cause is here:
https://github.com/ncalc/ncalc/blob/08994e2c366d4f2621aeea461c0e12ac1448dbf4/src/NCalc/Expression.cs#L205-L232
I think it would be better to throw a new `EvaluationException` using the overload `EvaluationException(string message, Exception innerException)` to add the original exception as inner exception. In the unit test, the inner exception could identify the cause with `DeclaringType == typeof(System.DateTimeParse)` rather than testing for the message.
Would you accept a PR for that?

Btw.: `publish-nuget.yml` should update the depreciated `actions/checkout@v2` with `actions/checkout@v4`.

axunonb:
Closes #120

### DIFF
--- a/src/NCalc/Expression.cs
+++ b/src/NCalc/Expression.cs
@@ -208,7 +208,7 @@ public class Expression
         }
         catch(Exception ex)
         {
-            var message = new StringBuilder(ex.Message);
+            var message = new StringBuilder();
             if (errorListenerLexer.Errors.Count != 0)
             {
                 message.AppendLine();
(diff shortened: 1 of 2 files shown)


gumbarros:
LGTM, thanks!

PR #129, opened 2024-04-27
Update project to create unsigend and signet versions of NCalcSync

### LINKED ISSUE #25: Assembly signing
I can't update to this new version, because i need signed assemblies (using net472). 

I need to copy that assembly to the GAC, because i use it in a SharePoint on-premise project.

axunonb:
Closes  #25

### Reasoning behind publishing both unsigned and strongly signed assemblies of NCalc

1. **Accessibility**: By providing both signed and unsigned versions, you cater to a broader audience. Some users or organizations may have strict policies that require signed assemblies, while others may not have such requirements and prefer unsigned versions. Offering both increases the accessibility of *NCalc* to users with varying security needs.

 2. **Compatibility**: Some users may be constrained by their development environment or dependencies, which may require either signed or unsigned assemblies. Providing separate packages allows users to choose the version that seamlessly integrates with their existing setup, potentially attracting more users who otherwise might have been unable to use *NCalc*.

 3. **Low Overhead**: The overhead of maintaining two versions of the assembly is minimal compared to the benefits it provides. The effort required to build and publish both versions is low, especially considering the potential increase in user base and the flexibility it offers to users.

 ### What's included in this PR

1. **NCalc.sln**: Add new configuration `SignedRelease`.

2. **NCalc.csproj**: 
 
    * Add new configuration `SignedRelease`, while keeping existing workflow for `Release` and `Debug` configurations.
    * With `SignedRelease`, the assembly will be signed with a strong name key file. For transparency, the strong name key file is included in the repository. This key file is used to sign the assembly during the build process. The key file is not password-protected and is intended for use by NCalc contributors.
    * `Description`s contain a hint for the signed or unsigned nuget package respectively.
    * Remove references to `net35` which is not a target framework.

3. **publish-nuget.yml**:

    * Update obsolete workflow actions to newer versions.

    * Add a new workflow for publishing the signed version of the assembly to NuGet. The workflow is triggered when a new release is created. It builds the project in the `SignedRelease` configuration and publishes the signed assembly to NuGet as `NCalcSync.signed` package.

4. **README.md**:
    Add a nuget bage with a link to the signed package.


gumbarros:
@yallie I don't know about assembly signing but LGTM

yallie:
Thanks for the PR!
Everything looks good to me.

What kind of backwards compatibility the signed version of the assembly would break?
Is there any downside of a signed assembly vs unsigned one?
Why would anyone prefer to keep using the unsigned version?

Thanks!

src/NCalc/NCalc.csproj
```diff
@@ -1,51 +1,73 @@
...
+
+NCalc can parse any expression and evaluate the result, including static or dynamic parameters and custom functions.
+
+This package contains the unsigned assembly 'NCalc'. For the stronly signed assembly 'NCalc.signed', see the NCalcSync.signed package.
+		</Description>
+	</PropertyGroup>
+
+	<!-- Strongly Signed Assembly Configuration -->
+	<PropertyGroup Condition=" '$(Configuration)' == 'SignedRelease' ">
+		<AssemblyName>NCalc.signed</AssemblyName>
```
yallie, line 38: Is it necessary to change the assembly name?

axunonb: Good point. Users will decide for the signed or the unsigned package, but never for both. So a collision of assembly names is impossible in practice.
The PR is updated, so the assembly name is NCalc.dll for both unsigned and signed versions.

axunonb:
> What kind of backwards compatibility the signed version of the assembly would break?

The signed version is new and exists in parallel to the unsigned one. So there's nothing I can imagine, that would break.

> Is there any downside of a signed assembly vs unsigned one?

The downside is that once an assembly is signed, it cannot be modified without invalidating the signature. This means that if you need to update or patch the assembly, you’ll need to sign it again. In case of NCalc, publishing nuget packages, you will publish a new release instead of a patch.

Why would anyone prefer to keep using the unsigned version?

* Anyone who has an unsigned assembly infrastructure, will keep it, no matter which frameworks they are using.

* Projects targeting .NET 5.0 and later versions **only (!)**, can load unsigned assemblies from signed ones. This is because the handling of trust boundaries and security policies has evolved in these .NET versions. They introduce more modern security features and mechanisms for handling different trust levels. :NET versions before v5.0 [will throw runtime exceptions](https://stackoverflow.com/questions/61273217/net-distributing-a-signed-strong-named-library-that-depends-on-an-unsigned) when loading unsigned assemblies from a signed one.

yallie:
Hi @axunonb, thanks for your reply.

>The signed version is new and exists in parallel to the unsigned one. So there's nothing I can imagine, that would break.

I'm sorry I should have been more specific in my question...

Do we really need a new nuget package?
What if we just sign the upcoming version of NCalc and keep the same NCalcSync package?
Can it break anything for the users of the unsigned assembly version when they upgrade?
Assembly binding redirections in the config files, something like that perhaps?

I would prefer just to sign the assembly and drop support for the unsigned one.
If it doesn't break any workflows, of course... What do you think?

axunonb:
> Do we really need a new NuGet package?

I did some further investigations and came across [a StackOverflow article about "signing an unsigned assembly"](https://stackoverflow.com/questions/1669139/signing-an-unsigned-assembly) covering these questions.

> Assembly binding redirections in the config files, something like that perhaps?

Yes, indeed, this will require modifications.

My conclusion for "new NuGet package" is **yes**, if we keep the major version, because switching to "signed only" may have implications for existing projects. Then, we should also emphasize: If your project creates unsigned assemblies, stick to using the unsigned `NCalc` version. Users with a "signed" project will have to use the signed `NCalc` anyway.

Publishing "signed only" is not a breaking change in the traditional sense (like changing the API), but it can still potentially cause disruptions. Going this way, a new major version could be the right choice to signal a "heads up" besides the release notes.

yallie:
Ok @axunonb, thanks a lot for your explanations and for the PR!

axunonb:
I created the signed nuget packages from the v3.12.0 release (commit https://github.com/ncalc/ncalc/commit/08994e2c366d4f2621aeea461c0e12ac1448dbf4) to publish the first signed version.
Files renamed to *.zip
[NCalcSync.signed.3.12.0.snupkg.zip](https://github.com/ncalc/ncalc/files/15202402/NCalcSync.signed.3.12.0.snupkg.zip)
[NCalcSync.signed.3.12.0.nupkg.zip](https://github.com/ncalc/ncalc/files/15202403/NCalcSync.signed.3.12.0.nupkg.zip)

axunonb:
@yallie What do think - could we publish [the packages](https://github.com/ncalc/ncalc/pull/129#issuecomment-2093325062) for v3.12.0 on NuGet?

yallie:
>could we publish https://github.com/ncalc/ncalc/pull/129#issuecomment-2093325062 for v3.12.0 on NuGet?

Yes, absolutely. Thank you for building the packages!
I've added you as an owner of a signed nuget package: https://www.nuget.org/packages/NCalcSync.signed
I'll figure out how to update NCalcSync API token so that we can push new releases from github.

If everything's fine, the new package version should appear on the readme badge in a few minutes.
Hope the next release will automatically publish the strong-name version.

Thanks a lot!

PR #132, opened 2024-05-07
Small code style changes

gumbarros:
Nothing affect runtime except a [small "peanut butter" adding `sealed` at a internal class](https://devblogs.microsoft.com/dotnet/performance-improvements-in-net-6/#peanut-butter).

PR #134, opened 2024-05-07
Fixed `BitwiseXOr` behavior

### LINKED ISSUE #133: `SerializationVisitor` Erroneously Serializes `BitwiseXOr` as `~`
The `LogicalExpression` type's `ToString()` method uses the `SerializationVisitor` type to construct the string representation of the expression tree. Unfortuntely, there is a defect in the code in `SerializationVisitor.cs`, line 98:

```cs
case BinaryExpressionType.BitwiseXOr:
    Result.Append("~ ");
    break;
```

The symbol for `BitwiseXOr` is `^`, but the serializer method returns `~`. Apparently this was confused with the unary `BitwiseNot` operator.

As a result, a `LogicalExpression` parsed from e.g. `1 ^ 2` will return `1 ~ 2` from `.ToString()`. And since this is not syntactically valid, trying to compile that resulting expression throws an expectable parsing error.

The good news is this is a super easy fix, the `~` just needs to be replaced with a `^` on that line of code.

gumbarros:
This closes #133

### DIFF
--- a/src/NCalc/Domain/SerializationVisitor.cs
+++ b/src/NCalc/Domain/SerializationVisitor.cs
@@ -95,7 +95,7 @@ public class SerializationVisitor : LogicalExpressionVisitor
                 break;
 
             case BinaryExpressionType.BitwiseXOr:
-                Result.Append("~ ");
+                Result.Append("^ ");
                 break;
 
             case BinaryExpressionType.LeftShift:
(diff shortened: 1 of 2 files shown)


yallie:
Looks like it conflicts with PR #132 because the fixtures had a lot of indentation changes.

gumbarros:
@yallie  fixed

PR #146, opened 2024-05-22
Refactored to use new LeftAssociative and Unary parsers

gumbarros:
@sebastienros all testing are passing, great job. Just `.Compile` is missing


Bykiev:
Parlot is updated, this PR can be updated too

Issue #151, opened 2024-05-22
Support function effects outside of expression


slobber:
I want create a function to check a value whether large than 10 three times. 

The Expression will be `Repeat([value] > 10, 3)`.

```
 var e = new Expression("Repeat([value] > 10, 3)");
 
 var times = new Dictionary<string, int>() { };

 e.EvaluateFunction += (name, args, id) =>
 {
     if (name == "Repeat")
     {
         var t = (int)args.Parameters[1].Evaluate() - 1;
         var r = (bool)args.Parameters[0].Evaluate();
         if (r && id != null)
         {
             if (!times.ContainsKey(id))
             {
                 times[id] = t;
             }
             else
             {
                 times[id]--;
             }
         }

         args.Result =  r && times[id] == 0;
     }
 };
 e.Parameters["value"] = 9;

 Assert.Equal(false, e.Evaluate());
 
 e.Parameters["value"] = 11;
 Assert.Equal(false, e.Evaluate());
 e.Parameters["value"] = 12;
 Assert.Equal(false, e.Evaluate()); 
 e.Parameters["value"] = 13;
 Assert.Equal(true, e.Evaluate()); 
```

gmcchessney:
We had a similar requirement (maintaining state within a unique function instance) and got around the limitation by pre-processing the expression. In our use-case we modified the function names to be unique, but you could also inject an "id" parameter into the functions. Here's an example repo with parameter injection, just take a look in NCalc.Play:

https://github.com/gmcchessney/ncalc/tree/unique-function-id-workaround-151

PR #152, opened 2024-05-22
Add an argument "id" in function handler to support the function effect.

slobber:
The "id" argument will identify the unique function instance, and make it possible to change the values outside the function. #151


gumbarros:
LGTM, @Bykiev what you think

Bykiev:
I don't like this idea, this logic should be handled inside `EvaluateFunction` event by creating `N` expressions

slobber:
I just want to make it possible that each function instance can save some internal state, and when the same expression is called again, the internal state of the function can participate in the calculation of the function result.

gumbarros:
I also don't see problem in the Id as it's optional. So, it's impossible to resolve this problem manually and this PR is a must have?

slobber:
Consider follow expression:
```
// Repeat is a function that will return true if the condition (arg 0) is true in (arg 1) times
Repeat([value1] > 3, 2) && Repeat([value2] > 10, 3)
```
each function call should have an independent state of itself. With `id` argument I can implement the function like this:

```
 var times = new Dictionary<string, int>() { };

 e.EvaluateFunction += (name, args, id) =>
 {
     if (name == "Repeat")
     {
         var r = (bool)args.Parameters[0].Evaluate();
         var t = (int)args.Parameters[1].Evaluate() - 1;
         if (r && id != null)
         {
           times[id] = !times.ContainsKey(id) ? t : times[id] - 1;
         }

         args.Result =  r && times[id] <= 0;
     }
 };
```
Is it possible implement this feature without `id`?

gumbarros:
```cs
 var times = new Dictionary<string, int>() { };
 var executionId = Guid.NewGuid();
 e.EvaluateFunction += (name, args) =>
 {
     if (name == "Repeat")
     {
         var r = (bool)args.Parameters[0].Evaluate();
         var t = (int)args.Parameters[1].Evaluate() - 1;
         if (r && id != null)
         {
           times[executionId] = !times.ContainsKey(id) ? t : times[executionId] - 1;
         }

         args.Result =  r && times[executionId] <= 0;
         executionId = Guid.NewGuid();
     }
 };
 ```
 
 Is this possible?

**Comment by slobber** [comment]
> ```cs
>  var times = new Dictionary<string, int>() { };
>  var executionId = Guid.NewGuid();
>  e.EvaluateFunction += (name, args) =>
>  {
>      if (name == "Repeat")
>      {
>          var r = (bool)args.Parameters[0].Evaluate();
>          var t = (int)args.Parameters[1].Evaluate() - 1;
>          if (r && id != null)
>          {
>            times[executionId] = !times.ContainsKey(id) ? t : times[executionId] - 1;
>          }
> 
>          args.Result =  r && times[executionId] <= 0;
>          executionId = Guid.NewGuid();
>      }
>  };
> ```
> 
> Is this possible?

No，the function call will generate a new id everytime. so whatever how many times the function call, the result is always `false`
In my solution, the id was set value when function instance created.

The test case below shows what i'm talking about. 
```
var e = new Expression("Repeat([value1] > 3, 2) && Repeat([value2] > 10, 3)");

e.Parameters["value1"] = 2;
e.Parameters["value2"] = 12;

Assert.Equal(false, e.Evaluate()); // value1 > 3 == false, value2 > 10 == true (once), result is false

e.Parameters["value1"] = 11;

Assert.Equal(false, e.Evaluate());  // value1 > 3 == true (once), value2 > 10 == true (twice), result is still false

Assert.Equal(true, e.Evaluate()); // value1 > 3 == true (twice), value2 > 10 == true (3 times), result is true
```

**Comment by Bykiev** [comment]
Good day, I'll dig into this case later, don't like current solution

**Comment by gmcchessney** [comment]
I commented on the issue with a workaround. We ran into this limitation a while back and worked around it by preprocessing the expression.

**Comment by slobber** [comment]
One more thing, in the above example, "Repeat ([Value1]> 3, 2) && Repeat ([Value2]> 10, 3)"
It will not run as expected, because logic operator "&&" will ignore the right expression, so we should change the formula to:
```
And(Repeat ([Value1]> 3, 2), Repeat ([Value2]> 10, 3))
```
and add a custom function:
```
if (name == "And")
{
  var left = (bool)args.Parameters[0].Evaluate();
  var right = (bool)args.Parameters[1].Evaluate();
  args.Result = left && right;
}
```

**Comment by gumbarros** [comment]
Good night, still confused in the discussion, this will enter 4.1 ?

**Comment by Bykiev** [comment]
As there is a warkaround and I still have no time to dig into this issue, I'd so maybe in v4.2. Or you could merge if you think it's ok

**Comment by gumbarros** [comment]
image

**Comment by Bykiev** [comment]
@slobber, I've created a sample, seems it should solve your issue without any code modification. Can you please check it?

```
internal class Program
{
    private static Dictionary<string, int> times = new Dictionary<string, int>() { };

    static void Main(string[] args)
    {
        var expr = new Expression("Repeat([value] > 10, 3)", CultureInfo.InvariantCulture);

        string guid = Guid.NewGuid().ToString();
        RegisterExpr(expr, guid);

        expr.Parameters["value"] = 9;
        Console.WriteLine(expr.Evaluate());

        expr.Parameters["value"] = 11;
        Console.WriteLine(expr.Evaluate());

        expr.Parameters["value"] = 12;
        Console.WriteLine(expr.Evaluate());

        expr.Parameters["value"] = 13;
        Console.WriteLine(expr.Evaluate());

        Console.ReadLine();
    }

    public static void RegisterExpr(Expression exp, string id)
    {
        exp.EvaluateFunction += delegate (string name, FunctionArgs args)
        {
            RegisterFunctions(name, args, id);
        };
    }

    private static void RegisterFunctions(string name, FunctionArgs args, string id)
    {
        if (name.Equals("Repeat", StringComparison.InvariantCultureIgnoreCase))
        {
            var t = (int)args.Parameters[1].Evaluate() - 1;
            var r = (bool)args.Parameters[0].Evaluate();

            if (r && id != null)
            {
                if (!times.ContainsKey(id))
                {
                    times[id] = t;
                }
                else
                {
                    times[id]--;
                }
            }

            args.Result = r && times[id] == 0;
        }
    }
}
```

**Comment by slobber** [comment]
@Bykiev ，Thanks. But it doesn't match my requirement. Think about this expression "And(Repeat ([Value1]> 3, 2), Repeat ([Value2]> 10, 3))"，in your solution, one expression only would have a guid. But this expression should have two guid for each Repeat function call.

**Comment by Bykiev** [comment]
We've already made a lot of breaking changes and I think one more is not a problem if needed. I still think it can be solved without any code  modification, will research your example at the evening.

**Comment by gumbarros** [comment]
@slobber v5 solves your problem with no breaking changes:
https://github.com/ncalc/ncalc/pull/254/commits/2cb0f0034c39f86de3c7ad745897f0b97426c0ce

**Comment by Bykiev** [comment]
@slobber, what about a class variable?

```
public class Calculator
{
    public Dictionary<string, int> d = new();
    private string id = Guid.NewGuid().ToString();
    private readonly Expression _expr;

    public Calculator(Expression expr)
    {
        _expr = expr;
        _expr.EvaluateFunction += delegate (string name, FunctionArgs args)
        {
            if (name == "SecretOperation")
            {
                if (id != null)
                {
                    if (!d.ContainsKey(id))
                    {
                        d[id] = 3;
                    }
                    else
                    {
                        d[id]--;
                    }
                }

                args.Result = (int)args.Parameters[0].Evaluate() + (int)args.Parameters[1].Evaluate();
            }
        };
    }

    public object Calculate()
    {
        id = Guid.NewGuid().ToString();
        return _expr.Evaluate();
    }
}
```

```
var expr = new Expression("SecretOperation(3, 6) + SecretOperation(1, 2)");
var calculator = new Calculator(expr);

Console.WriteLine(calculator.Calculate());
Console.WriteLine(calculator.Calculate());
Console.WriteLine(calculator.d.FirstOrDefault().Value);

Console.ReadLine();
```

**Review by Bykiev — CHANGES_REQUESTED** [review]
I'm approving it as I don't see any workarounds. Conflicts should be resolved before merge, the docs should be updated too

**Comment by gumbarros** [comment]
> I'm approving it as I don't see any workarounds. Conflicts should be resolved before merge, the docs should be updated too
We should check with @slobber if https://github.com/ncalc/ncalc/pull/254/commits/2cb0f0034c39f86de3c7ad745897f0b97426c0ce solves the problem, so this PR is not needed and no breaking changes are made.

**Comment by slobber** [comment]
> > I'm approving it as I don't see any workarounds. Conflicts should be resolved before merge, the docs should be updated too
> 
> We should check with @slobber if [2cb0f00](https://github.com/ncalc/ncalc/commit/2cb0f0034c39f86de3c7ad745897f0b97426c0ce) solves the problem, so this PR is not needed and no breaking changes are made.

This test case does not reflect my actual requirements.

I think the test case may be like follow code.
```
    [Fact]
    public void ExpressionShouldEvaluateCustomFunctionsWithEffects()
    {
        var e = new Expression("And(Repeat([value] > 10, 3), Repeat([value] < 15, 2)");
        e.Functions["And"] = (args, context) => {
            var left = (bool)args.Parameters[0].Evaluate();
            var right = (bool)args.Parameters[1].Evaluate();
            args.Result = left && right;
        };
        const string id = "Repeat";
        
        var times = new Dictionary<string, int>();
        e.Functions[id] = (args, context) =>
        {

            var t = (int)args[1].Evaluate() - 1;
            var r = (bool)args[0].Evaluate();
            if (r)
            {
                if (!times.ContainsKey(id))   // There're two Repeat function in the expression, each call should have its identity, each call should have independent state inside (times[id])
                {
                    times[id] = t;
                }
                else
                {
                    times[id]--;
                }
            }
            return r && times[id] == 0;
        };
        e.Parameters["value"] = 9;    // [value] > 10 (times["[value] > 10, 3"] == 3), [value] <15 == true (times["[value] < 15, 3"] == 1)

        Assert.Equal(false, e.Evaluate());

        e.Parameters["value"] = 11;   // [value] > 10 == true (times["[value] > 10, 3"] == 2), [value] <15 == true (times["[value] < 15, 3"] == 0)
        Assert.Equal(false, e.Evaluate());
        e.Parameters["value"] = 12;   // [value] > 10 == true (times["[value] > 10, 3"] == 1), [value] <15 == true (times["[value] < 15, 3"] == 0)
        Assert.Equal(false, e.Evaluate()); 
        e.Parameters["value"] = 13;
        Assert.Equal(true, e.Evaluate());   // [value] > 10 == true (times["[value] > 10, 3"] == 0), [value] <15 == true (times["[value] < 15, 3"] == 0)，the expression return true
        Assert.Single(times);
    }

```

**Comment by gumbarros** [comment]
> > > I'm approving it as I don't see any workarounds. Conflicts should be resolved before merge, the docs should be updated too
> > 
> > 
> > We should check with @slobber if [2cb0f00](https://github.com/ncalc/ncalc/commit/2cb0f0034c39f86de3c7ad745897f0b97426c0ce) solves the problem, so this PR is not needed and no breaking changes are made.
> 
> This test case does not reflect my actual requirements.
> 
> I think the test case may be like follow code.
> 
> ```cs
>     [Fact]
>     public void ExpressionShouldEvaluateCustomFunctionsWithEffects()
>     {
>         var e = new Expression("And(Repeat([value] > 10, 3), Repeat([value] < 15, 2)");
>         e.Functions["And"] = (args, context) => {
>             var left = (bool)args.Parameters[0].Evaluate();
>             var right = (bool)args.Parameters[1].Evaluate();
>             args.Result = left && right;
>         };
>         const string id = "Repeat";
>         
>         var times = new Dictionary<string, int>();
>         e.Functions[id] = (args, context) =>
>         {
> 
>             var t = (int)args[1].Evaluate() - 1;
>             var r = (bool)args[0].Evaluate();
>             if (r)
>             {
>                 if (!times.ContainsKey(id))   // There're two Repeat function in the expression, each call should have its identity, each call should have independent state inside (times[id])
>                 {
>                     times[id] = t;
>                 }
>                 else
>                 {
>                     times[id]--;
>                 }
>             }
>             return r && times[id] == 0;
>         };
>         e.Parameters["value"] = 9;    // [value] > 10 == false (times["[value] > 10, 3"] == 3), [value] <15 == true (times["[value] < 15, 3"] == 1)
> 
>         Assert.Equal(false, e.Evaluate());
> 
>         e.Parameters["value"] = 11;   // [value] > 10 == true (times["[value] > 10, 3"] == 2), [value] <15 == true (times["[value] < 15, 3"] == 0)
>         Assert.Equal(false, e.Evaluate());
>         e.Parameters["value"] = 12;   // [value] > 10 == true (times["[value] > 10, 3"] == 1), [value] <15 == true (times["[value] < 15, 3"] == 0)
>         Assert.Equal(false, e.Evaluate()); 
>         e.Parameters["value"] = 13;
>         Assert.Equal(true, e.Evaluate());   // [value] > 10 == true (times["[value] > 10, 3"] == 0), [value] <15 == true (times["[value] < 15, 3"] == 0)，the expression return true
>         Assert.Single(times);
>     }
> ```

Now I understand. This is also a reason to have event handlers along the lambda dictionaries.

To prevent a breaking change what about:
```cs
public class FunctionArgs(Guid id, Expression[] parameters) : EventArgs
{
    private object? _result;

    public object? Result
    {
        get => _result;
        set
        {
            _result = value;
            HasResult = true;
        }
    }
    
    public Guid Id { get; } = id; //<--HERE

    public Expression[] Parameters { get; } = parameters;
    
    public bool HasResult { get;  private set; }

    public object?[] EvaluateParameters()
    {
        var values = new object?[Parameters.Length];
        for (var i = 0; i < values.Length; i++)
        {
            values[i] = Parameters[i].Evaluate();
        }

        return values;
    }
}
```

gumbarros:
Super seeded by #266

PR #157, opened 2024-05-22
Add missing Ncalc.Benchmarks project to sln

Bykiev:
Add missing Ncalc.Benchmarks project to sln

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -19,6 +19,8 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
       - name: Dotnet Setup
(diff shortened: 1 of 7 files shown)


gumbarros:
Before merging, I think we should study on how to make the Benchmark work with both .csproj and NuGet version. I 100% agree to the benchmark be add the .sln, but if we compile the solution in debug or release and not SignedRelease, it cause errors (specifically at CI)

Bykiev:
I've made it draft and opened an issue https://github.com/dotnet/BenchmarkDotNet/issues/2574

Bykiev:
As InProcess toolchain is not supporting benchmarking with NuGet package we can run it on CI when building separate executables

gumbarros:
Superseded by #220

PR #161, opened 2024-05-23
Add support for number decimal separator based on specified culture

Bykiev:
This PR adds support for number decimal separator based on specified culture (dot or comma).

Stuck with getting the `ParseContext` to get the culture from `LogicalExpressionParserContext`, help is wanted


gumbarros:
I think `ParseContext` is only available inside `.Then`, one way should be expecting at the parser OneOf() all of the CultureInfo separators

Bykiev:
Right now I see only 1 solution: add a public `CultureInfo` property to `LogicalExpressionParser`

sebastienros:
You could definitely have a `CultureInfo` in the `LogicalExpressionParser`. How it's done right now in NCalc it's due to the existing samples in the Parlot repository. Parsers usually don't vary and are thread-safe so having a static instance makes sense most of the time.

However you can create different instances, and some parts of the parser can still be static, just use the ctor and pass it an argument. You just need to ensure that you don't create a new instance every time you parse.

Unless each expression evaluation can take a `CultureInfo`, in that case the parser needs to use it during parse time to make decisions, and it's still possible. You say only `Then` has it, but we can add it to other parsers if that makes sense. There is a `Switch` parser that can be used to return different parsers dynamically for this exact scenario, and it accepts a `ParseContext`.

In the lambda try to return another static parser such that again each evaluation doesn't create a new `Parser<T>` instance.

test/NCalc.Tests/CustomCultureTests.cs
```diff
@@ -65,36 +65,23 @@ void ExecuteTest(string formula, object expectedValue)
...
 
-            var e = new Expression("[a]<2.0", CultureInfo.InvariantCulture);
-            e.Parameters["a"] = "1.7";
-            Assert.Equal(true, e.Evaluate());
-        }
-        finally
-        {
-            Thread.CurrentThread.CurrentCulture = originalCulture;
-        }
+        var e = new Expression("[a]<2,0", culture);
```
sebastienros, line 83: You will need to add something like `Round(3,123)` and see how it behaves.
I don't think allowing any decimal separator anywhere is sustainable.

Bykiev:
```
var culture = (CultureInfo)CultureInfo.InvariantCulture.Clone();
culture.NumberFormat.NumberDecimalSeparator = ",";

var e = new Expression("Round(3,123,0)", culture);
Assert.Equal(3d, e.Evaluate());
```

Test is passing

Bykiev: I agree, it's not looking good, it'll be better to replace NCalc parameter delimiter to semicolon, instead of comma

sebastienros: > Assert.Equal(3d, e.Evaluate());

Is it `0` because it just discarded the "second" argument? What about `If([Pi]>3.14,3.14,3.13)` but with commas ;)

Bykiev: > > Assert.Equal(3d, e.Evaluate());
> 
> Is it `0` because it just discarded the "second" argument? What about `If([Pi]>3.14,3.14,3.13)` but with commas ;)

```
var culture = (CultureInfo)CultureInfo.InvariantCulture.Clone();
culture.NumberFormat.NumberDecimalSeparator = ",";

var e = new Expression("if([Pi] > 3,14, 3,14, 3,13)", culture);
e.Parameters["Pi"] = Math.PI;
Assert.Equal(3.14d, e.Evaluate());
```
Test is passing

sebastienros: Cheater ;) 

I assume that could work if you just passed the decimal separator and an argument separator which are distinct, like `,` and `;`. As long as the user decides, their problem.

I personally don't care, just explaining all the issues I can see.

Bykiev: I love your idea to make it possible to use `,` and `;` as argument separator. Thank you for your inputs!

gumbarros:
I will try to check this for the next version after `async`

Bykiev:
> I will try to check this for the next version after `async`

It's seems not working as expected, still needs a lot of research

Bykiev:
> @Bykiev Parlot is now at v1 with Number parsers, we can give a try again

It seems with cultures where comma is decimal separator it's not possible to use comma as parameter separator anymore, only semicolons...

gumbarros:
Too much conflicts and no direction in how to do this without risks. Closing for now.

PR #162, opened 2024-05-23
Added support for using semicolon as argument separator

Bykiev:
Added support for using semicolon as argument separator. This PR is a part of using culture-specific decimal number separator. If comma is decimal separator it'll be more readable if semicolon is used (e.g. "Round(1,23;1)")

### DIFF
--- a/docs/articles/functions.md
+++ b/docs/articles/functions.md
@@ -32,3 +32,5 @@ It also includes other general purpose ones.
 |in	|Returns whether an element is in a set of values.	|in(1 + 1, 1, 2, 3)	|true|
 |if	|Returns a value based on a condition.	|if(3 % 2 = 1, 'value is true', 'value is false')	|'value is true'|
 |ifs    |Returns a value based on evaluating a number of conditions, returning a default if none are true. | ifs(foo > 50, "bar", foo > 75, "baz", "quux") | if foo is between 50 and 75 "bar", foo greater than 75 "baz", otherwise "quux" |  
+
+You can use comma (,) or semicolon (;) for argmument separator.
--- a/src/NCalc/Parser/LogicalExpressionParser.cs
+++ b/src/NCalc/Parser/LogicalExpressionParser.cs
@@ -36,7 +36,7 @@ public static class LogicalExpressionParser
          *                  | "(" expression ")" ;
          *
          * function       => Identifier "(" arguments ")"
-         * arguments      => expression ( "," expression )*
+         * arguments      => expression ( ("," | ";") expression )*
          */
         // The Deferred helper creates a parser that can be referenced by others before it is defined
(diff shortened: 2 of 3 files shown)


gumbarros:
Interesting, there is not situation that a semicolon would break a previously existing expression right? Also if a user don't specify the culture, it will behave like this right?

```
func(1,2;3) -> My parameters are "1", "2" and "3" //en-us
func(1,2;3) -> My parameters are "1,2" and 3 //pt-br
```

Bykiev:
> Interesting, there is not situation that a semicolon would break a previously existing expression right? 

All tests are passing, so, should be fine

>Also if a user don't specify the culture, it will behave like this right?
> 
> ```
> func(1,2;3) -> My parameters are "1", "2" and "3" //en-us
> func(1,2;3) -> My parameters are "1,2" and 3 //pt-br
> ```
Interesting, need to research it...

Bykiev:
Can't reproduce the issue with Russian culture. Here is my test case:

```
var e = new Expression("func(1,2;3)", CultureInfo.CurrentCulture);
e.EvaluateFunction += delegate (string name, Handlers.FunctionArgs args)
{
    Assert.Equal(3, args.Parameters.Length);
};
```

Bykiev:
Sorry for misunderstood, Yes, if comma is decimal separator it will be treated as 2 parameters, with point - as 3 parameters

gumbarros:
> Sorry for misunderstood, Yes, if comma is decimal separator it will be treated as 2 parameters, with point - as 3 parameters. That's why the preferred way is using semicolon as argument separator for comma decimal separator.
> 
> `func(1,2,3)` will have 2 parameters too with comma decimal separator (1.2 as first argument and 3 as second).
> 
> But with this PR comma is treated as argument separator. Using culture-specific decimal separator is another PR. So, this PR doesn't contain any breaking changes and just adds possibility to use comma or semicolon as argument separator.

Nice, I was just confirming the behavior. Will be useful at #118

Bykiev:
> > Sorry for misunderstood, Yes, if comma is decimal separator it will be treated as 2 parameters, with point - as 3 parameters. That's why the preferred way is using semicolon as argument separator for comma decimal separator.
> > `func(1,2,3)` will have 2 parameters too with comma decimal separator (1.2 as first argument and 3 as second).
> > But with this PR comma is treated as argument separator. Using culture-specific decimal separator is another PR. So, this PR doesn't contain any breaking changes and just adds possibility to use comma or semicolon as argument separator.
> 
> Nice, I was just confirming the behavior. Will be useful at #118

Internally we're using curly braces too, but as it's not officially supported by NCalc we're replacing them with square brackets. I'm ready to add support for curly braces if you're agreed.

PR #166, opened 2024-05-24
Fix handling invalid token

### LINKED ISSUE #150: Regression when updating to v4. - Invalid tokens are silently ignored
**Describe the bug**
When expression with invalid token was evaluated in version 3.12 , error was returned. In v4 it returns the value without error. 

**Example Expression**
-42a

Expected result: exception
Actual result: -42

In this example, there is no parameter called "a" defined etc, it's just an arbitrary symbol.

Bykiev:
This PR fixes using invalid tokens inside expression.

Closes #150

Though, all tests are passing I don't like the implementation. For some reason such expression is not working:

`.AndSkip(Literals.Identifier().Error("Invalid token in expression"))`

@gumbarros, do you have ideas why it's not working?

### DIFF
--- a/src/NCalc/Parser/LogicalExpressionParser.cs
+++ b/src/NCalc/Parser/LogicalExpressionParser.cs
@@ -1,4 +1,5 @@
 using NCalc.Domain;
+using NCalc.Exceptions;
 using Parlot.Fluent;
 using static Parlot.Fluent.Parsers;
 using Identifier = NCalc.Domain.Identifier;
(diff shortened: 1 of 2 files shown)


Bykiev:
@sebastienros, can you please check why does this one is not working

`AndSkip(Literals.Identifier().Error("Invalid token in expression"))`

but such is working:

`AndSkip(Not(Literals.Identifier()).ElseError("Invalid token in expression"))`

Looks like a Parlot bug for me...

gumbarros:
Good morning, sorry, Idk, I tried to do nearly the exact same thing as you before working on DI

Bykiev:
> Good morning, sorry, Idk, I tried to do nearly the exact same thing as you before working on DI

Good day, looks like a Parlot bug, I'll wait for Sébastien's response

sebastienros:
@Bykiev what is the issue in Parlot, here is a test I added locally:

```c#
        [Fact]
        public void ElseErrorShouldWorkAfterAndSkipNot()
        {
            var parser = Terms.Char('+').AndSkip(Not(Literals.Identifier()).ElseError("Invalid"));

            Assert.True(parser.TryParse("+123", out var result, out var error));
            Assert.Equal('+', result);
            Assert.Null(error);            

            Assert.False(parser.TryParse("+abc", out result, out error));
            Assert.Equal(default, result);
            Assert.Equal("Invalid", error.Message);            
        }
```

Expect a `+` then something that won't match an identifier, otherwise fails. Please provide a failing test if you can, that will help.

sebastienros:
... will try the one that you actually said was failing ;)

sebastienros:
Test for the other case, which will always fail: if Identifier is matched then Error will kick in and fail the parsing. If Identifier doesn't match then AndSkip will fail.

```c#
        [Fact]
        public void ErrorShouldFailWhenPreviousMatched()
        {
            var parser = Terms.Char('+').AndSkip(Literals.Identifier().Error("Invalid"));

            Assert.False(parser.TryParse("+123", out var result, out var error));
            Assert.Equal('\0', result);
            Assert.Null(error);            

            Assert.False(parser.TryParse("+abc", out result, out error));
            Assert.Equal(default, result);
            Assert.Equal("Invalid", error.Message); 
        }
```

sebastienros:
My experience is that you can sprinkle `ElseError` in places you want to provide an error message when the previously expected parser didn't succeed. So not to change the outcome, but provide an explanation in the error. There are some examples in Fluid.

Bykiev:
Thank you, got the idea!

Bykiev:
@gumbarros, ready for review

gumbarros:
We should improve exception handling to tell the invalid token position, this will be useful in large expressions. But I vote in create a separated issue and PR.

https://github.com/sebastienros/parlot/blob/main/src/Parlot/ParseException.cs

PR #169, opened 2024-05-25
Add support fo curly braces as alternative to square brackets

Bykiev:
Add support fo curly braces as alternative to square brackets


gumbarros:
@Bykiev thanks for the feature, LGTM.

gumbarros (approved):

PR #170, opened 2024-05-25
Remove `Logical_Expression_From_Cache_Should_Be_The_Same` to work with `dotnet test`

gumbarros:


gumbarros:
Help wanted before releasing 4.1, the `Logical_Expression_From_Cache_Should_Be_The_Same` works at the IDE, but sometimes work, sometimes fail at `dotnet test`

gumbarros:
image

gumbarros:
image

And after some tries, it fails. I think we should not test `IMemoryCache` implementation. `TryGetValue` is out of NCalc scope I think, and like the method name says, it will try, it's not 100% it works.

Issue #168, opened 2024-05-24
Do not convert external function name to lower case if ignoreCase is specified


Bykiev:
Right now when the external function is called and `ignoreCase` is used the function name is converted to lower case. I think we shouldn't convert it to any case and pass the original name to external. In most cases `ToUpperInvariant()` should be used in external implementation (the same as for built-in functions). It'll be a breaking change again, so, need to be discussed.

https://github.com/ncalc/ncalc/blob/abd34129c13b04a76a308e7127d95253c5844908/src/NCalc/Visitors/EvaluationVisitor.cs#L286

gumbarros:
100% agree, I event added a warning at my docs because of this weird behavior

![image](https://github.com/ncalc/ncalc/assets/52143624/ebbbe030-8579-433b-a11c-ff00329e8821)

PR #179, opened 2024-05-26
Do not convert external function name to lower case when ExpressionOptions.IgnoreCase option is used

Bykiev:
This PR removes convertion of external function name to lower case when ExpressionOptions.IgnoreCase option is used. The original function name will be passed to custom function evalution. You should handle this logic in your application.

It's a breaking change.

Closes [#168](https://github.com/ncalc/ncalc/issues/168)


gumbarros:
Nice work, should this be v5 or v4.2? `master` version.json file is pointing to a future v4.2

gumbarros (approved):

Bykiev:
Don't know exactly, but I think we can release it in v4.2. I believe it has a low impact as in most cases it's already handled in external application

gumbarros:
Agree, as this don't affect source code and just runtime. Please update `CHANGELOG.MD` and we are ready.

Bykiev:
Thank you, done

Issue #175, opened 2024-05-25
Better exception handling


gumbarros:
**Is your feature request related to a problem? Please describe.**
Yes, the old parser returns info with the token position

**Describe the solution you'd like**
Parlot also supports this, we should just add the info to `NCalcParserException` (because we support any parser)

**Describe alternatives you've considered**
None

**Additional context**
None

Issue #172, opened 2024-05-25
NCalcAsync should be merged at this repo.


gumbarros:
Probably a v5 milestone. 

[ncalc-async](https://github.com/ncalc/ncalc-async) is 6 months behind, and every new feat or fix needs to be manually added to the other repo. 

The best way to solve this problem would be:

```mermaid
graph TD;
    NCalcSync-->NCalc;
    NCalcAsync--> NCalc;
```

gumbarros:
Also a design flaw (I think)

Only `EvaluationVisitor` should be async and probably 80% of the codebase can be re-used. Because all code runs sync and just return `Task.CompletedTask`. The big advantage of NCalcAsync is to do CPU/IO bound work inside event handlers of `IEvaluationVisitor`.

gumbarros:
```cs

public interface IEvaluationVisitorContract : ILogicalExpressionVisitor
{
    public event EvaluateFunctionHandler? EvaluateFunction;
    public event EvaluateParameterHandler? EvaluateParameter;
    
    ExpressionContext Context { get; set; }
    
    public Dictionary<string, object?> Parameters { get; set; }
}
public interface IEvaluationVisitor : IEvaluationVisitorContract
{    
    public object? Result { get; }
}

public interface IAsyncEvaluationVisitor : IEvaluationVisitorContract
{
    public Task? Result { get; }
}
```

This would solve all the problems in theory.

gumbarros:
Also, if this works (looks very promising for me), we can in theory have a single Assembly with both sync and async, but idk what everyone thinks on this, and the name NCalc is already taken.

ThomasHambach:
Thumbs up on my end. That would definitely streamline the work we have to put into merging the features/fixes! I think keeping the package name as is would be fine as in the codebase you are already referring to `Ncalc` and not `NcalcSync`. We just need to add the proper info on the `NcalcAsync` package.

gumbarros:
> me), we can in theory have a single Assembly with both sync and async, but idk what everyone thinks on this, and the name NCalc is already taken.

I think we can just work with:

NCalc -> Assembly not published to NuGet, (but if someone wants we can), with all shared business logic.
NCalcSync -> Assembly with `Expression` and `EvaluationVisitor`
NCalcAsync -> Assembly with `AsyncExpression` and `AsyncEvaluationVisitor`

ThomasHambach:
[...]

PR #182, opened 2024-05-26
Move `Type` logic to `TypeHelper` and improve exception handling.

gumbarros:
Small refactors.

- Improve exception handling (#175) with two new exceptions (`NCalcParameterNotDefinedException`, `NCalcFunctionNotFoundException`) and use `NCalcEvaluationException` instead of `ArgumentException`
- Starting to remove some logic from `EvalutationVisitor` so it can be used at `async` (#172 )


Bykiev (approved):
LGTM

Bykiev:
It seems to be a breaking change too

gumbarros:
> It seems to be a breaking change too

Nice catch, a runtime breaking change, because if someone is expecting a `ArgumentException` it will not work.

I will update the changelog before merging

PR #184, opened 2024-05-27
Add support for using null with operators

Bykiev:
This PR adds support for using null with operators. In this case null value will be returned instead of InvalidOperationException. This is a breaking change


Bykiev:
Related discussion: https://github.com/ncalc/ncalc/discussions/183

gumbarros:
If someone types 2 + null, the equivalent is 2 or null? Because 2 + 0 = 2

Bykiev:
> If someone types 2 + null, the equivalent is 2 or null? Because 2 + 0 = 2

Any operation with null will return null value

gumbarros:
> > If someone types 2 + null, the equivalent is 2 or null? Because 2 + 0 = 2
> 
> Any operation with null will return null value

image


Sorry, makes total sense

PR #191, opened 2024-05-28
Fix OverflowException

### LINKED ISSUE #190: Regression bug - OverflowException with double values
I've found that in `LogicalExpressionParser` the result variable type is `decimal`, but I think it should be `double` and only when `ExpressionOptions.DecimalAsDefault` (`UseDecimalsAsDefault`) specified it should be converted to decimal. 

https://github.com/ncalc/ncalc/blob/e6c8c5f47932810a9f2f00397a95d31d058b4722/src/NCalc/Parser/LogicalExpressionParser.cs#L88-L95

Because of this such expression cause overflow:

```
var expr = new Expression($"Floor({double.MaxValue.ToString(CultureInfo.InvariantCulture)})");
var res = expr.Evaluate();
```

Seems to be a regression bug since 4.0

Bykiev:
Closes #190

### DIFF
--- a/src/NCalc/Parser/LogicalExpressionParser.cs
+++ b/src/NCalc/Parser/LogicalExpressionParser.cs
@@ -70,29 +70,29 @@ public static class LogicalExpressionParser
                 long? decimalPart = x.Item2;
                 long? exponentPart = x.Item3;
 
-                decimal result = integralValue;
+                double result = integralValue;
 
                 // decimal part?
                 if (decimalPart != null && decimalPart.Value != 0)
                 {
                     var digits = Math.Floor(Math.Log10(decimalPart.Value) + 1);
-                    result += decimalPart.Value / (decimal)Math.Pow(10, digits);
+                    result += decimalPart.Value / Math.Pow(10, digits);
                 }
 
                 // exponent part?
                 if (exponentPart != null)
                 {
-                    result *= (decimal)Math.Pow(10, exponentPart.Value);
+                    result *= Math.Pow(10, exponentPart.Value);
                 }
 
                 if (ctx is LogicalExpressionParserContext { UseDecimalsAsDefault: true })
                 {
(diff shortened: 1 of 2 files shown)


Bykiev:
For some reason with .NET Framework the result is different...

gumbarros:
> For some reason with .NET Framework the result is different...
> 
> **Upd.** With .NET8 the result is `1.7976931348623157`, but with .NET Framework it's `1.79769313486232`. Need to investigate it...

Strange, is there a issue related this at dotnet/runtime?

Bykiev:
> Strange, is there a issue related this at dotnet/runtime?

Sorry, changed your message by mistake. Seems to be related https://learn.microsoft.com/en-us/dotnet/core/compatibility/core-libraries/7.0/max-precision-numeric-format-strings

Bykiev:
https://devblogs.microsoft.com/dotnet/floating-point-parsing-and-formatting-improvements-in-net-core-3-0/

Bykiev:
It seems to be related how the Parlot parses the expression. I don't know if we can do anything with it.
@sebastienros, can you please look at it?

Bykiev:
There is also a precision loss issue that never happend in v3:

```
var expr = new Expression($"Floor(12e+100)");
var res = expr.Evaluate();

Assert.Equal(Math.Floor(12e+100), res);
```

```
Expected: 1,1999999999999999E+101
Actual:   1,2000000000000001E+101
```

Bykiev:
Though it doesn't related to this PR, but this expression is still failing both in .NET Framework and .NET:

```
var expr = new Expression($"Floor(12e+100)");
var res = expr.Evaluate();

Assert.Equal(Math.Floor(12e+100), res);
```

Seems due to how floating point values are stored in memory, I think it can be resolved only by using `BigDecimal` implementation. 

I think it should be another PR.

PR #188, opened 2024-05-28
Added `LambdaExpressionVisitor` and new `ExpressionOptions` from NCalc2

gumbarros:
This closes #185 
This closes #190 
This closes #189 

Credits to @sklose for the original implementation.

New ExpressionOptions:
`AllowBooleanCalculation`
`OrdinalStringComparer`

Co-authored-by: Sebastian Klose
[mail@sklose.com](mailto:mail@sklose.com)

Co-authored-by: Andrew Bykiev
[Bykiev](https://github.com/Bykiev)

PR #193, opened 2024-05-29
Fix double precision issue

Bykiev:
This PR fixes double precision


gumbarros:
Nice, but this PR is also bringing everything from #188 ?

PR #207, opened 2024-06-05
Added `async` support

### LINKED ISSUE #172: NCalcAsync should be merged at this repo.

gumbarros:
This closes #172


gumbarros:
Next steps:
- More unit tests (focused on evaluation, parsing can be totally ignored)
- Should we split in two assemblies? NCalcSync and NCalcAsync?
- Can we re-use even more code?

gumbarros:
I would also like some inputs if we should have two assemblies or not.

Points for two assemblies:
- Separation of concerns
- Naming makes more sense at NuGet
- Unnecessary instances of `EvaluationVisitor `are not created when using `async` and vice-versa.

Negative points for two assemblies:
- More hard to maintain CI/CD pipelines (I think)
- Possibility of dependency hell

Points for single assembly:
- More easy to maintain
- No dependency hell

Negative points for single assembly:
- Name does not make sense

Bykiev:
My vote for 2 separate assemblies as `NCalcSync` name can be confusing. We can always publish both of them with the same version even if there is no any changes.

gumbarros:
> My vote for 2 separate assemblies as `NCalcSync` name can be confusing. We can always publish both of them with the same version even if there is no any changes.

Agree, the naming is the most critical point for me.


```mermaid
graph TD;
    NCalc.Sync-->NCalc;
    NCalc.Async--> NCalc;
    NCalc.DependencyInjection-->NCalc.Async
    NCalc.DependencyInjection-->NCalc.Sync
    NCalc.Antlr-->NCalc.DependencyInjection
    NCalc.MemoryCache-->NCalc.DependencyInjection
```

What do you think of this structure?

Bykiev:
Looking good for me

gumbarros:
> Looking good for me

Nice, this had been a busy week for me, I will try to finish this at weekend/next week.

gumbarros:
@Bykiev Ready for review.

The only thing that changed at the plan is the base package is named `NCalc.Core` because `NCalc` is certainly already taken. Do you have another name suggestion? The .csproj for us is `NCalc`, just the NuGet (and it's not intended for direct consumage)

gumbarros:
Btw, I vote in not manually changing the changelog, GitHub already generates a changelog at release and I can copy and paste it

Bykiev:
I like the Ncalc.Core name.
Sorry, I'm on the phone, can't review you changes carefully today, only tomorrow evening at UTC +3.

Not related, but the performance of last versions is getting lower, now Parlot is only about 50% faster than Antlr. I think we should rethink the Benchmark project and instead add a diff bench Parlot vs Antlr, plus a benchmark with lambda compilation

gumbarros:
> I like the Ncalc.Core name. Sorry, I'm on the phone, can't review your changes carefully today, only tomorrow evening at UTC +3.
> 
> Not related, but the performance of last versions is getting lower, now Parlot is only about 50% faster than Antlr. I think we should rethink the Benchmark project and instead add a diff bench Parlot vs Antlr, plus a benchmark with lambda compilation

Agree, probably the more rules we add at the parser, more slow it will become. After this PR, I can create another PR comparing both implementations of LogicalExpressionFactory. Makes more sense than comparing old NuGets

Bykiev:
PR is welcome :)

gumbarros:
> PR is welcome :)

Nice, after this PR is merged I will re-add the Benchmark project and add two benchmarks:

Antlr LogicalExpressionFactory vs Parlot LogicalExpressionFactory

Evaluate vs ToLambda

Bykiev (approved):
Good job, thanks!

PR #225, opened 2024-06-10
Inline `TypeHelper.IsReal`

gumbarros:
// * Summary *

BenchmarkDotNet v0.13.12, Windows 11 (10.0.22631.3672/23H2/2023Update/SunValley3)
13th Gen Intel Core i5-1335U, 1 CPU, 12 logical and 10 physical cores
.NET SDK 8.0.300
  [Host]     : .NET 8.0.5 (8.0.524.21615), X64 RyuJIT AVX2
  DefaultJob : .NET 8.0.5 (8.0.524.21615), X64 RyuJIT AVX2

| Method          | Mean      | Error     | StdDev    | Median    | Rank | Gen0   | Allocated |
|---------------- |----------:|----------:|----------:|----------:|-----:|-------:|----------:|
| AggresiveInline | 0.0003 ns | 0.0011 ns | 0.0014 ns | 0.0000 ns |    1 |      - |         - |
| NoInline        | 2.5937 ns | 0.0539 ns | 0.0504 ns | 2.5791 ns |    2 | 0.0038 |      24 B |

I think this is the only method in the codebase that can be inlined. (small and no control flow)

### DIFF
--- a/src/NCalc.Core/Helpers/TypeHelper.cs
+++ b/src/NCalc.Core/Helpers/TypeHelper.cs
@@ -1,4 +1,5 @@
 ﻿using System.Collections.Frozen;
+using System.Runtime.CompilerServices;
 
 namespace NCalc.Helpers;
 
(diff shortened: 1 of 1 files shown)


Bykiev (approved):
Though I'm not a big fan of such manual compiler optimizations, but this one looks good

PR #227, opened 2024-06-10
Fixed not operator behavior

### LINKED ISSUE #226: Breaking Change in 4.2.1 - not<nospace>(true) throws UnknownFunction
**Describe the bug**

The behavior has changed - expression with no space between not and a brace () throws a unknown function not.
Previos behavior was  'not ( true )' return false.
 

**Example Expression**
 'not( true )'  throws
 'not ( true )'  return false
 
 Should be added to Breaking Changes, or could not be handled as Operator instead function ?

gumbarros:
This closes #226

### DIFF
--- a/src/NCalc.Core/Parser/LogicalExpressionParser.cs
+++ b/src/NCalc.Core/Parser/LogicalExpressionParser.cs
@@ -145,8 +145,7 @@ public static class LogicalExpressionParser
         var identifier = Terms.Identifier();
 
         var not = OneOf(
-            Terms.Text("NOT", true)
-                .AndSkip(OneOf(Literals.WhiteSpace(), openParen.Then(e => new TextSpan(e.ToString())))),
+            Terms.Text("NOT", true).AndSkip(OneOf(Literals.WhiteSpace().Or(Not(AnyCharBefore(openParen))))),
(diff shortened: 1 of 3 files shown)


gumbarros:
Don't know if it's the best way, but better than a custom function I think

Bykiev:
Can you please add `not(true)` failing case in the test

gumbarros:
> Can you please add `not(true)` failing case in the test

Weird why this fails ☠️, I think creating the custom function will do the job

Bykiev:
Sure, but it'll be a breaking change as functions are case-sensitive, but operators not

gumbarros:
> Sure, but it'll be a breaking change as functions are case-sensitive, but operators not

At ANTLR this works, weird, I see no problem in a function, but there should be a way at the parser

Bykiev:
Definitly it's possible to fix it on parser side. As I can see after this it's getting `true)` and because of this it's not treated as group expression

Bykiev:
I think it can be handled in a custom way in `Then` flow parser

Bykiev (approved):

gumbarros:
@Bykiev thanks for the help

PR #237, opened 2024-06-11
Add support UInt64 for binary operators

### LINKED ISSUE #41: Value was either too large or too small for a UInt16.
I'm trying to do binary operators ^ | on integer values representing flags to only keep the desired values. My values are actually 64bits (ulong), but NCalc seems to do all its internal numeric computation on 16bits. Any way around that?

Bykiev:
This PR adds support for using UInt64 for binary operators. This is a sort of breaking change, so, will be better to release in v4.4

Closes #41


gumbarros:
Why is this a breaking change? The return value of the exp?

gumbarros:
I will merge this tonight, before I need to change `version.json` to 4.3.1 and release your small fix

PR #246, opened 2024-06-12
Require braces to be closed by a brace of the same type

### LINKED ISSUE #244: Consider requiring braces to be closed by a brace of the same type
**Is your feature request related to a problem? Please describe.**
A change was made recently to allow `{` and `}` to enclose identifiers. This will cause problems for many of our existing expressions that contain these characters within an identifier. For example:
```
[{Diagnostic}.Data]
```
The parser will treat `[{Diagnostic}` as the identifier, and the rest as invalid syntax.

**Describe the solution you'd like**
Square braces should only be closed by square braces, and curly braces should only be closed by curly braces. Within a section of square braces, curly braces can be treated no differently than any other character, and vice-versa. This is similar to how strings work in JavaScript.

**Describe alternatives you've considered**
We can continue to use the ANTLR parser until / unless this is resolved.

gumbarros:
This closes #244


Bykiev (approved):
LGTM

PR #253, opened 2024-06-13
Fix parsing fractional zero

### LINKED ISSUE #252: Error parsing expressions when only zero after point
**Describe the bug**
Error parsing expressions when only zero in fractional part

**Example Expression**
`.0`

Bykiev:
Closes #252


Bykiev:
Just found `1.` case is not passing now and there is no tests...

Issue #107, opened 2024-02-22
Repeated Evaluate() calls cause increased amount custom function resolving.


AnttiPeko:
Simple test to count how many times custom function is evaluated. I would expect function to be evaluated once per call to Evaluate(), but amount increases every time so at 10th evaluate call function will get resolved 10 times.

```
using NCalc;

var counter = 0;
var totalCounter = 0;

var expression = new Expression("MyFunc()");

expression.EvaluateFunction += Expression_EvaluateFunction;

for (var i = 0; i < 10; i++)
{
    counter = 0;
    var result=expression.Evaluate();
    Console.WriteLine($"{counter} {totalCounter}");
}


void Expression_EvaluateFunction(string name, FunctionArgs args)
{
    if (name != "MyFunc") return;
    args.Result = 1;
    counter++;
    totalCounter++;
}
```
`<PackageReference Include="NCalcSync" Version="3.10.0" />`

Results to
```
1 1
2 3
3 6
4 10
5 15
6 21
7 28
8 36
9 45
10 55
```

PR #254, opened 2024-06-14
Added `IDictionary<string,ExpressionFunction>` and `IDictionary<string,ExpressionParameter>` support

### LINKED ISSUE #229: Refactor expression arguments
The functions are passed as delegates on every expression instantiation. This is sub-optimal, instead a user should be able to register the methods once and then reuse this registration.

The way I do it in [Fluid](https://github.com/sebastienros/fluid) is by defining a global object `FluidOptions` that can be shared across expression instances. You could then create other specific option instances by copy like `new FluidOptions(baseOptions)` if you want to inherit from common options but change a single thing. This could then be used for all the options of an expression, including culture, arguments, ...

The instance should be thread-safe for read operations so it can be made static an shared.

gumbarros:
This closes #229


gumbarros:
Breaking changes: Yes

Docs coming soon

Bykiev:
I don't see any real performance changes by this change, maybe we need some more benchmarks?

gumbarros:
> I don't see any real performance changes, maybe we need some more benchmarks?

It's not about performance but readability and a more easy public API. Check the unit tests to see the before and after.

Buut, I think performance should come as a dessert, because should be more faster not using events, I will do a benchmark later

gumbarros:
@Bykiev mixed results on the benchmark, slightly slow, but less memory footprint, but I think the performance impact is negligent.
```cs
using System.Collections.Frozen;
using BenchmarkDotNet.Running;
using BenchmarkDotNet.Attributes;
using BenchmarkDotNet.Order;

namespace NCalc.Benchmarks;

public class ExampleArgs : EventArgs
{
    public string Name { get; set; }
    public object Result { get; set; }
}
public class ExampleExpression
{
    public Dictionary<string, Func<object>> LambdaDictionary;
    public FrozenDictionary<string, Func<object>> FrozenLambdaDictionary;
    public event EventHandler<ExampleArgs> MyEvent;
        
    public object EvaluateFromDictionary()
    {
        return LambdaDictionary["SecretOperation"]();
    }
    
    public object EvaluateFromFrozenDictionary()
    {
        return FrozenLambdaDictionary["SecretOperation"]();
    }

    public object EvaluateEvent()
    {
        var args = new ExampleArgs()
        {
            Name = "SecretOperation"
        };
        MyEvent?.Invoke(this, args);
        return args.Result;
    }
}


[RankColumn]
[CategoriesColumn]
[MemoryDiagnoser]
[Orderer(SummaryOrderPolicy.FastestToSlowest)]
public class EventHandlerVsLambdaBenchmark
{
    private ExampleExpression _exampleExpression;

    [GlobalSetup]
    public void Setup()
    {
        _exampleExpression = new ExampleExpression();
        _exampleExpression.LambdaDictionary = new Dictionary<string, Func<object>>
        {
            { "SecretOperation", () => 42 }
        };
        _exampleExpression.FrozenLambdaDictionary = new Dictionary<string, Func<object>>
        {
            { "SecretOperation", () => 42 }
        }.ToFrozenDictionary();
        _exampleExpression.MyEvent += (sender, args) =>
        {
            if (args.Name == "SecretOperation")
                args.Result = 42;
        };
    }

    [Benchmark]
    public object EventHandler()
    {
        return _exampleExpression.EvaluateEvent();
    }

    [Benchmark]
    public object LambdaDictionary()
    {
        return _exampleExpression.EvaluateFromDictionary();

    }

    [Benchmark]
    public object FrozenLambdaDictionary()
    {
        return _exampleExpression.EvaluateFromFrozenDictionary();
    }
}

public class Program
{
    public static void Main(string[] args)
    {
        _ = BenchmarkRunner.Run<EventHandlerVsLambdaBenchmark>();
    }
}
```

| Method                 | Mean      | Error     | StdDev    | Rank | Gen0   | Allocated |
|----------------------- |----------:|----------:|----------:|-----:|-------:|----------:|
| FrozenLambdaDictionary |  4.987 ns | 0.0769 ns | 0.0720 ns |    1 | 0.0038 |      24 B |
| EventHandler           |  6.330 ns | 0.0824 ns | 0.0770 ns |    2 | 0.0089 |      56 B |
| LambdaDictionary       | 10.876 ns | 0.0618 ns | 0.0548 ns |    3 | 0.0038 |      24 B |

gumbarros:
Expected breaking changes:

Functions and parameters are refactored for a more cleaner public API.

# Functions:

v4:
```cs
expression.EvaluateFunction += delegate(string name, FunctionArgs args)
    {
        if (name == "SecretOperation")
            args.Result = (int)args.Parameters[0].Evaluate() + (int)args.Parameters[1].Evaluate();
    };
```

v5:
```cs
expression.Functions["SecretOperation"] = (args, context) => {
    return (int)args[0].Evaluate() + (int)args[1].Evaluate();
};
```

# Dynamic Parameters:

v4:
```cs
expression.EvaluateParameter += delegate(string name, ParameterArgs args)
  {
    if (name == "Pi")
        args.Result = 3.14;
  };
```

v5:
```cs
expression.DynamicParameters["Pi"] = (context) => {
    Console.WriteLine("I'm evaluating π!");
    return 3.14;
};
``` 
## Other small changes:
- Removed `IEvaluationVisitor`, please use `IEvaluationService`
- Removed `IParameterExtractionVisitor`
- `ILogicalExpression` factory now expects a `LogicalExpressionOptions` object instead of a `ExpressionContext`
- At `NCalc.Async`, `ExpressionContext` is now `AsyncExpressionContext`
- Removed `ExpressionOptions.IgnoreCase`, please check [the docs.](https://ncalc.github.io/ncalc/articles/case_sensitivity.html)

Bykiev:
Wow, I need some more time for review these changes.
But what will be with v4, as we have recently released it, should we still maintain both v4 and v5? Due to logic breaking changes the migration will be not fast and many of user still will be using v4 for a long time. I still expect some parser bugs in v4.

gumbarros:
> Wow, I need some more time for review these changes. But what will be with v4, as we have recently released it, should we still maintain both v4 and v5? Due to logic breaking changes the migration will be not fast and many of users still will be using v4 for a long time. I still expect some parser bugs in v4.

We can release v5 at preview mode and still fix bugs at v4.x in a separate branch for a X time. But I don't know if we have the strength to maitain two versions.

Another option would be re-adding the event handlers, marking them with `ObsoleteAttribute` and use them to populate the `ExpressionContext`

Bykiev:
> Another option would be re-adding the event handlers, marking them with `ObsoleteAttribute` and use them to populate the `ExpressionContext`
This seems to be a better solution as we really can't maintain both versions due to lack of resources.

gumbarros:
> > Another option would be re-adding the event handlers, marking them with `ObsoleteAttribute` and use them to populate the `ExpressionContext`
> 
> This seems to be a better solution as we really can't maintain both versions due to lack of resources.

I changed my mind, the effort to re-add the event handlers is not worth I think, the codebase will also became more hard to maintain analyzing where I should put again the event handlers.

Everything is so nice in this version because the entire context of the expression can be tracked from `ExpressionContext`, re-adding event handlers would cause a lot of headaches like issue #107

The migration effort for consumers should be small, literally copying the content of a function, moving the name to a dictionary and returning the value instead of setting.

Bykiev:
I'd say let's try to maintain both v4 and v5 then. The only expected changes in v4 are related to Parlot parser. I believe it can be easily merged in both branches. What do you think?

gumbarros:
> I'd say let's try to maintain both v4 and v5 then. The only expected changes in v4 are related to Parlot parser. I believe it can be easily merged in both branches. What do you think?

I think is the only way then, `master` should be v5 and v4 fixes can be merged at the latest v4 branch.

Only problem would be how to have two docs.

Bykiev:
I'm still on the phone, so, maybe I missed smth. In our app the user is creating custom formulas and typical formula is looking this way: 
`MyCustomFunc({ParamId; ParameterTypeId; ObjectId})`
With the event handlers we have parsed the parameters by semicolon, retrieved the value and then function called. How to achieve it with dynamic parameters? Should we compile the expression first to retrieve the parameters?

gumbarros:
> I'm still on the phone, so, maybe I missed smth. In our app the user is creating custom formulas and typical formula is looking this way: `MyCustomFunc({ParamId; ParameterTypeId; ObjectId})` With the event handlers we have parsed the parameters by semicolon, retrieved the value and then function called. How to achieve it with dynamic parameters? Should we compile the expression first to retrieve the parameters?

After you are in the PC can you share a snippet of `expression.EvaluateFunctoin += ...`?

Bykiev:
We have a formula `ChildSum({ParamId, ParamTypeId})` when registering custom functions and parameters we're passing additional parameters (this seems to be not supported now)

```
  private void RegisterFunctions(Object obj, Expression exp, DateTime currDt)
  {
      exp.EvaluateParameter += delegate (string name, ParameterArgs args)
      {
          args.Result = RegisterExpressionParameters(name, currDt, obj);
      };

      exp.EvaluateFunction += delegate (string name, FunctionArgs args)
      {
          RegisterExpressionFunctions(name, currDt, args, obj);
      };
  }
```

The `RegisterExpressionFunctions` code is similar to this for `ChildSum` custom function:
 
```
double sum = 0;
var children = GetChildren(obj);
foreach(var child in children) {
  var childExpr = new Expression(args.Parameters[0].LogicalExpression, _expressionOptions, CultureInfo.InvariantCulture);
  
  RegisterFunctions(rObject, childExpr);
  var result = GetResult(childExpr);
  sum += result.HasValue ? result.Value > 0 ? 1 : 0 : 0;
}

args.Result = sum;
```

The `RegisterExpressionParameters` logic:
```

var param = name.Split(new char[] { ',' });

string pType = param[0];
int? pId = Convert.ToInt32(param[1]);

return GetValue(pId, pType, obj.Id, currDt);
```

gumbarros:
> We have a formula `ChildSum({ParamId, ParamTypeId})` when registering custom functions and parameters we're passing additional parameters (this seems to be not supported now)
> 
> ```
>   private void RegisterFunctions(Object obj, Expression exp, DateTime currDt)
>   {
>       exp.EvaluateParameter += delegate (string name, ParameterArgs args)
>       {
>           args.Result = RegisterExpressionParameters(name, currDt, obj);
>       };
> 
>       exp.EvaluateFunction += delegate (string name, FunctionArgs args)
>       {
>           RegisterExpressionFunctions(name, currDt, args, obj);
>       };
>   }
> ```
> 
> The `RegisterExpressionFunctions` code is similar to this for `ChildSum` custom function:
> 
> ```
> double sum = 0;
> var children = GetChildren(obj);
> foreach(var child in children) {
>   var childExpr = new Expression(args.Parameters[0].LogicalExpression, _expressionOptions, CultureInfo.InvariantCulture);
>   
>   RegisterFunctions(rObject, childExpr);
>   var result = GetResult(childExpr);
>   sum += result.HasValue ? result.Value > 0 ? 1 : 0 : 0;
> }
> 
> args.Result = sum;
> ```
> 
> The `RegisterExpressionParameters` logic:
> 
> ```
> 
> var param = name.Split(new char[] { ',' });
> 
> string pType = param[0];
> int? pId = Convert.ToInt32(param[1]);
> 
> return GetValue(pId, pType, obj.Id, currDt);
> ```

Ok, I have no minimal idea in how to refactor this to a Dictionary of delegates, maybe this is possible because the delegate signature also contains `Expression[]` as args.
You conviced me to think more in how to still have the event handlers, I'm re-adding them inside the `ExpressionContext` and re-adding the unit tests to make sure everything is ok, soon I will commit.

gumbarros:
TLDR for the incoming commit:
```cs
    /// <summary>
    /// Event triggered to handle function evaluation.
    /// </summary>
    public event EvaluateFunctionHandler EvaluateFunction
    {
        add => Context.EvaluateFunction += value;
        remove => Context.EvaluateFunction -= value;
    }
    
    /// <summary>
    /// Event triggered to handle parameter evaluation.
    /// </summary>
    public event EvaluateParameterHandler EvaluateParameter
    {
        add => Context.EvaluateParameter += value;
        remove => Context.EvaluateParameter -= value;
    }
    ```

**Comment by gumbarros** [comment]
@Bykiev docs and unit tests for the good old way re-added. If you figure out how to solve your cause using `IDictionary` we can mark the event handlers with `ObsoleteAttribute` and add to the docs a migration tutorial.

**Comment by Bykiev** [comment]
> @Bykiev docs and unit tests for the good old way re-added. If you figure out how to solve your cause using `IDictionary` we can mark the event handlers with `ObsoleteAttribute` and add to the docs a migration tutorial.

I believe it's possible but have no ideas for now how to implement it. Let's leave the event handlers for such cases.

**Review by Bykiev — APPROVED** [review]
LGTM

**Comment by gumbarros** [comment]
I will wait for @sebastienros and @yallie before merging

**Comment by gumbarros** [comment]
T

> I thought I had sent a comment from my phone but apparently it didn't work.
> 
> I think you should benchmark using `ValueTask` vs `Task`. This will most probably be faster and in that case it should use it.

Tks, `ValueTask` wins

BenchmarkDotNet v0.13.12, Ubuntu 24.04 LTS (Noble Numbat)
AMD Ryzen 5 3600, 1 CPU, 12 logical and 6 physical cores
.NET SDK 8.0.300
  [Host]     : .NET 8.0.5 (8.0.524.21615), X64 RyuJIT AVX2
  DefaultJob : .NET 8.0.5 (8.0.524.21615), X64 RyuJIT AVX2


| Method                          | Mean     | Error    | StdDev   | Rank | Gen0   | Allocated |
|-------------------------------- |---------:|---------:|---------:|-----:|-------:|----------:|
| EvaluateFromValueTaskDictionary | 57.31 ns | 0.360 ns | 0.337 ns |    1 | 0.0057 |      96 B |
| EvaluateFromTaskDictionary      | 75.11 ns | 1.361 ns | 1.206 ns |    2 | 0.0143 |     240 B |
https://gist.github.com/gumbarros/51107469d656602a4bdb9f483e8f91e3

**Review by Bykiev — APPROVED** [review]

**Comment by gumbarros** [comment]
Good morning @yallie can you review when possible 👀?

**Comment by Bykiev** [comment]
Hope Alexey is ok, let's merge it into master.
Only one question: as there is no any breaking changes, should we really release it as v5 and not v 4.4?

**Comment by yallie** [comment]
Hello guys!

Sorry for my late response, I can't do a proper review right now.
Let me just say I like new API a lot, looks much cleaner to me overall 👍

>old way:
>```cs
>expression.EvaluateFunction += delegate(string name, FunctionArgs args)
>   {
>        if (name == "SecretOperation")
>            args.Result = (int)args.Parameters[0].Evaluate() + (int)args.Parameters[1].Evaluate();
>    };
>```
>
>new way:
>```cs
>expression.Functions["SecretOperation"] = (args, context) => {
>    return (int)args[0].Evaluate() + (int)args[1].Evaluate();
>};
>```

One quick question: is function/parameter name somehow available in the new handlers?
Perhaps via the `context` variable?

**Review by yallie — APPROVED** [review]
Looks great to me! 🚀

**Comment by gumbarros** [comment]
> One quick question: is function/parameter name somehow available in the new handlers?

No, because the name is the key of the dictionary itself. Because of this and some other scenarios, the old way is still avaiable without breaking changes.

It would be possible to add to context, but when a user is creating a context the first time, would be strange the name property appearing and would be redundant because of the dictionary key.

**Comment by yallie** [comment]
>No, because the name is the key of the dictionary itself. Because of this and some other scenarios, the old way is still avaiable without breaking changes.

I'm asking because the dictionary key can be out of scope for the function handler, i.e:

```csharp
expression.Functions["SecretOperation"] = CompileFunctionScript(); // function handler doesn't have access to the dictionary key
```

yallie:
>It would be possible to add to context, but when a user is creating a context the first time, would be strange the name property appearing and would be redundant because of the dictionary key.

Oh, I see, the context argument is `ExpressionContext`, the container for functions, parameters, etc.
I thought it's more like local function evaluation context or something like that.
Of course it's not the proper place for the function name :)

But can we have function name in `FunctionArgs` and `AsyncFunctionArgs`?

gumbarros:
> > It would be possible to add to context, but when a user is creating a context the first time, would be strange the name property appearing and would be redundant because of the dictionary key.
> 
> Oh, I see, the context argument is `ExpressionContext`, the container for functions, parameters, etc. I thought it's more like local function evaluation context or something like that. Of course it's not the proper place for the function name :)
> 
> But can we have function name in `FunctionArgs` and `AsyncFunctionArgs`?

I think this is now related to #152, I suggested to add the Id of the function at FunctionArgs, but would be more appropriate to add the entire `Function` at `FunctionArgs` so it's possible to know both Name and Id

yallie:
>I suggested to add the Id of the function at FunctionArgs, but would be more appropriate to add the entire Function at FunctionArgs so it's possible to know both Name and Id.

Yes, `FunctionArgs` seems to be the best place for the function name and its internal state.

>At IDictionary<string, ExpressionFunction> and IDictionary<string, ExpressionParameter> it's impossible to know

That's my concern: a new cleaner way has great API, but doesn't provide the function name to the handler. So the new API is missing some features, so it can't completely replace legacy API.

gumbarros:
> > I suggested to add the Id of the function at FunctionArgs, but would be more appropriate to add the entire Function at FunctionArgs so it's possible to know both Name and Id.
> 
> Yes, `FunctionArgs` seems to be the best place for the function name and its internal state.
> 
> > At IDictionary<string, ExpressionFunction> and IDictionary<string, ExpressionParameter> it's impossible to know
> 
> That's my concern: a new cleaner way has great API, but doesn't provide the function name to the handler. If the new API is missing a feature, it can't completely replace legacy API.
> 
> UPD. Not sure whether it's important and are we going to support both APIs indefinitely... I guess we can merge as is :)

I vote to support both APIs, as they are handled at the same method anyway. About the func/param name, I don't see reason to pass the name if the person who is implementing the function, knows the name because the dict key is immutable. But I 100% agree in passing the name using the legacy API, because a single delegate can handle multiple functions (this already happens).

gumbarros:
I did a final commit before merging, reviews are welcome

src/NCalc.Sync/Visitors/EvaluationVisitor.cs
```diff
@@ -239,7 +234,7 @@ public void Visit(Identifier identifier)
     protected int CompareUsingMostPreciseType(object? a, object? b)
     {
         return TypeHelper.CompareUsingMostPreciseType(a,
-            b, new(Context.CultureInfo,
+            b, new TypeHelper.ComparasionOptions(Context.CultureInfo,
```
yallie, line 237: >TypeHelper.ComparasionOptions

A possible typo? Should be ComparisonOptions IMO.

gumbarros: It's because it's a scoped type inside TypeHelper. I accidentally ctrl+. here, but I will change soon

gumbarros:
@yallie check latest commit. Should be more easy now

gumbarros:
Ohh sorry, now I see what you said haha. Fixed at 6ce0105266e4623c1539f0735f6d9fd165b8887c

gumbarros:
`ExpressionContext` is now a record to allow the with keyword. At expression_context.md I explain better. What you all think?

src/NCalc.Core/Helpers/ComparasionOptions.cs
```diff
@@ -0,0 +1,15 @@
+namespace NCalc.Helpers;
+
+public record struct ComparasionOptions
```
yallie: Should be `ComparisonOptions`, is it?

yallie: Sorry! Somehow got outdated commit :(

yallie:
> `ExpressionContext` is now a record to allow the `with` keyword. At `expression_context.md` I explain better. 

I like it! 👍

Bykiev:
There is smth wrong with benchmarks, can you please check it?

gumbarros:
> There is smth wrong with benchmarks, can you please check it?

image
At performance?

Bykiev:
| Method                   | Mean     | Error    | StdDev   | Rank | Gen0   | Allocated |
|------------------------- |---------:|---------:|---------:|-----:|-------:|----------:|
| SimpleAntlrExpression    |       NA |       NA |       NA |    ? |     NA |        NA |
| AdvancedAntlrExpression  |       NA |       NA |       NA |    ? |     NA |        NA |
| SimpleParlotExpression   | 10.08 us | 0.018 us | 0.014 us |    1 |      - |   1.05 KB |
| AdvancedParlotExpression | 29.05 us | 0.178 us | 0.166 us |    2 | 0.0305 |   2.58 KB |

Benchmarks with issues:
  LogicalExpressionFactoryBenchmark.SimpleAntlrExpression: DefaultJob
  LogicalExpressionFactoryBenchmark.AdvancedAntlrExpression: DefaultJob

gumbarros:
At GH Actions:
 // * Summary *

BenchmarkDotNet v0.13.12, Ubuntu 22.04.4 LTS (Jammy Jellyfish)
.NET SDK 8.0.302
  [Host]     : .NET 8.0.6 (8.0.624.26715), X64 RyuJIT AVX2
  DefaultJob : .NET 8.0.6 (8.0.624.26715), X64 RyuJIT AVX2


| Method                   | Mean     | Error    | StdDev   | Rank | Gen0   | Allocated |
|------------------------- |---------:|---------:|---------:|-----:|-------:|----------:|
| SimpleAntlrExpression    |       NA |       NA |       NA |    ? |     NA |        NA |
| AdvancedAntlrExpression  |       NA |       NA |       NA |    ? |     NA |        NA |
| SimpleParlotExpression   | 10.08 us | 0.018 us | 0.014 us |    1 |      - |   1.05 KB |
| AdvancedParlotExpression | 29.05 us | 0.178 us | 0.166 us |    2 | 0.0305 |   2.58 KB |

At my machine:
// * Summary *

BenchmarkDotNet v0.13.12, Ubuntu 24.04 LTS (Noble Numbat)
AMD Ryzen 5 3600, 1 CPU, 12 logical and 6 physical cores
.NET SDK 8.0.105
  [Host]     : .NET 8.0.5 (8.0.524.21615), X64 RyuJIT AVX2
  DefaultJob : .NET 8.0.5 (8.0.524.21615), X64 RyuJIT AVX2


| Method                   | Mean      | Error     | StdDev    | Rank | Gen0   | Allocated |
|------------------------- |----------:|----------:|----------:|-----:|-------:|----------:|
| SimpleAntlrExpression    |        NA |        NA |        NA |    ? |     NA |        NA |
| AdvancedAntlrExpression  |        NA |        NA |        NA |    ? |     NA |        NA |
| SimpleParlotExpression   |  9.333 us | 0.1844 us | 0.2462 us |    1 | 0.1221 |   1.05 KB |
| AdvancedParlotExpression | 26.854 us | 0.3874 us | 0.3624 us |    2 | 0.3052 |   2.58 KB |


At GH Actions:
| Method                   | Mean           | Error       | StdDev      | Rank | Gen0   | Allocated |
|------------------------- |---------------:|------------:|------------:|-----:|-------:|----------:|
| LambdaWithoutCompilation |      0.3095 ns |   0.0037 ns |   0.0029 ns |    1 |      - |         - |
| Evaluate                 |    902.0826 ns |   9.4986 ns |   8.8850 ns |    2 | 0.0229 |    1976 B |
| LambdaWithCompilation    | 22,984.5348 ns | 239.1400 ns | 223.6917 ns |    3 |      - |    5000 B |

At my machine:
| Method                   | Mean           | Error       | StdDev      | Rank | Gen0   | Gen1   | Allocated |
|------------------------- |---------------:|------------:|------------:|-----:|-------:|-------:|----------:|
| LambdaWithoutCompilation |      0.5287 ns |   0.0340 ns |   0.0318 ns |    1 |      - |      - |         - |
| Evaluate                 |    928.9449 ns |  18.4743 ns |  26.4953 ns |    2 | 0.2346 |      - |    1976 B |
| LambdaWithCompilation    | 30,304.6888 ns | 580.8553 ns | 621.5086 ns |    3 | 0.5493 | 0.4883 |    5022 B |

gumbarros:
@Bykiev I think the benchmark is not so precise, sometimes my machine is faster and sometimes GH is faster.

gumbarros:
Ohhh, the NA at ANTLR 😅...

Fixed:
// * Summary *

BenchmarkDotNet v0.13.12, Ubuntu 24.04 LTS (Noble Numbat)
AMD Ryzen 5 3600, 1 CPU, 12 logical and 6 physical cores
.NET SDK 8.0.105
  [Host]     : .NET 8.0.5 (8.0.524.21615), X64 RyuJIT AVX2
  DefaultJob : .NET 8.0.5 (8.0.524.21615), X64 RyuJIT AVX2


| Method                   | Mean      | Error     | StdDev    | Rank | Gen0   | Gen1   | Allocated |
|------------------------- |----------:|----------:|----------:|-----:|-------:|-------:|----------:|
| SimpleParlotExpression   |  8.941 us | 0.0179 us | 0.0159 us |    1 | 0.1221 |      - |   1.05 KB |
| SimpleAntlrExpression    |  9.398 us | 0.0784 us | 0.0695 us |    2 | 1.6327 | 0.0610 |  13.34 KB |
| AdvancedParlotExpression | 26.500 us | 0.0432 us | 0.0361 us |    3 | 0.3052 |      - |   2.58 KB |
| AdvancedAntlrExpression  | 29.481 us | 0.0662 us | 0.0587 us |    4 | 4.6997 | 0.5493 |  38.39 KB |

gumbarros:
Also, created a C# proposal because of `ExpressionContext`
https://github.com/dotnet/csharplang/issues/8246

I don't like changing it to `record` so we can have the `with` keyword.

Bykiev (approved):

PR #255, opened 2024-06-14
Refactor MathHelper

Bykiev:
This PR refactors MathHelper code.

PR contains a breaking change: I dont know why, but for some reason for modulo operation this types are not allowed:

```
TypeCode.Single and TypeCode.Decimal,
TypeCode.Double and  TypeCode.Decimal
TypeCode.Decimal and TypeCode.Single, TypeCode.Double
```

And for subtract `TypeCode.Single` and `TypeCode.Decimal` is not allowed. Is any reasons for this behaviour?

### DIFF
--- a/src/NCalc.Core/Helpers/MathHelper.cs
+++ b/src/NCalc.Core/Helpers/MathHelper.cs
@@ -471,7 +471,7 @@ public static class MathHelper
             double d => ExecuteDoubleOperation(d, b, operatorName, func),
             decimal @decimal => ExecuteDecimalOperation(@decimal, b, operatorName, func),
             _ => throw new InvalidOperationException(
-                                $"Operator '+' not implemented for operands of types {a} and {b?.GetType().ToString() ?? "null"}"),
+                                $"Operator '{operatorName}' not implemented for operands of types {a} and {b?.GetType().ToString() ?? "null"}"),
         };
     }
 
@@ -523,7 +523,7 @@ public static class MathHelper
                                 $"Operator '{operatorName}' can't be applied to operands of types 'ushort' and 'bool'"),
             byte or sbyte or short or ushort or int or uint or long or ulong or float or double or decimal => func(left, right),
             _ => throw new InvalidOperationException(
-                                $"Operator '+' not implemented for types 'ushort' and {right?.GetType().ToString() ?? "null"}"),
+                                $"Operator '{operatorName}' not implemented for types 'ushort' and {right?.GetType().ToString() ?? "null"}"),
         };
     }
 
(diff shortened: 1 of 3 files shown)


Bykiev:
Please, pay attention that with dynamics the performance will be worse, but the readability was increased after refactoring. I vote for readability in this case.

Bykiev:
I did a benchmark func and method with aggressive inlining and the results a pretty similar, so, I don't think it's needed
```
BenchmarkDotNet v0.13.12, Windows 10 (10.0.19044.2965/21H2/November2021Update)
Intel Core i7-7700 CPU 3.60GHz (Kaby Lake), 1 CPU, 8 logical and 4 physical cores
.NET SDK 8.0.205
  [Host] : .NET 6.0.30 (6.0.3024.21525), X64 RyuJIT AVX2

Toolchain=InProcessEmitToolchain

| Method             | Mean     | Error    | StdDev   | Rank | Gen0   | Allocated |
|------------------- |---------:|---------:|---------:|-----:|-------:|----------:|
| Multiply           | 19.61 ns | 0.395 ns | 1.145 ns |    2 | 0.0172 |      72 B |
| MultiplyAggressive | 16.46 ns | 0.392 ns | 0.675 ns |    1 | 0.0172 |      72 B |

// * Warnings *
MultimodalDistribution
  TestAgressive.Multiply: Toolchain=InProcessEmitToolchain -> It seems that the distribution can have several modes (mValue = 3.07)
```

gumbarros:
Wow, the performance is regression is 100% worth for the more easy to maintain MathHelper, congrats.

Will just check on the pc better before merging

Bykiev:
> Wow, the performance is regression is 100% worth for the more easy to maintain MathHelper, congrats.
> 
> Will just check on the pc better before merging

Please, pay attention for the breaking changes. As I can see it's even just a bug as I don't see any reasons why this operation are not allowed

gumbarros:
> > Wow, the performance is regression is 100% worth for the more easy to maintain MathHelper, congrats.
> > Will just check on the pc better before merging
> 
> Please, pay attention for the breaking changes. As I can see it's even just a bug as I don't see any reasons why these operations are not allowed

The breaking change is now TypeCode.Single and TypeCode.Decimal are allowed? For me this is a fix right?

Bykiev:
> > > Wow, the performance is regression is 100% worth for the more easy to maintain MathHelper, congrats.
> > > Will just check on the pc better before merging
> > 
> > 
> > Please, pay attention for the breaking changes. As I can see it's even just a bug as I don't see any reasons why these operations are not allowed
> 
> The breaking change is now TypeCode.Single and TypeCode.Decimal are allowed? For me this is a fix right?

Yes and I think it's a bug. I've checked the NCalc2 code and it was supported too

gumbarros:
> > > > Wow, the performance is regression is 100% worth for the more easy to maintain MathHelper, congrats.
> > > > Will just check on the pc better before merging
> > > 
> > > 
> > > Please, pay attention for the breaking changes. As I can see it's even just a bug as I don't see any reasons why these operations are not allowed
> > 
> > 
> > The breaking change is now TypeCode.Single and TypeCode.Decimal are allowed? For me this is a fix right?
> 
> Yes and I think it's a bug. I've checked the NCalc2 code and it was supported too

It's a fix then, not a breaking change 😄

PR #256, opened 2024-06-14
Overflow protection

### LINKED ISSUE #206: Checked Math
Add CheckedMath from NCalc2, just this option is missing.

**Additional context**
Already implemented at `LambdaExpressionVisitor`, but `MathHelper` is not prepared. Not a priority for me, but if someone is interested, a PR is welcome :)

Bykiev:
This PR adds support for handling overflow with binary arithmetic operators (+,-,*,/). To handle overflow `ExpressionOptions.OverflowProtection` should be specified.

Closes #206

### DIFF
--- a/src/NCalc.Core/ExpressionEvaluator.cs
+++ b/src/NCalc.Core/ExpressionEvaluator.cs
@@ -41,7 +41,8 @@ public abstract class ExpressionEvaluator
     
     protected MathHelperOptions MathHelperOptions => new(Context.CultureInfo,
         Context.Options.HasFlag(ExpressionOptions.AllowBooleanCalculation),
-        Context.Options.HasFlag(ExpressionOptions.DecimalAsDefault));
+        Context.Options.HasFlag(ExpressionOptions.DecimalAsDefault),
+        Context.Options.HasFlag(ExpressionOptions.OverflowProtection));
 
     protected void ExecuteUnaryExpression(UnaryExpression expression)
     {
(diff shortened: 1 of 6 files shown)


Bykiev:
Overflow detection seems to be tricky...

Bykiev:
One note for reviewers: `float.MaxValue + 1` will not cause overflow, because 1 is too little to overflow. Looks strange, but it's ok

gumbarros:
I was generating v4.3.3 when you open the P.R for review ☠️.

I will review carefully tonight after work. Should this be at v5?

Bykiev:
No problem, v5 will be ok

Bykiev:
It seems we need docs versioning for v4 and v5

src/NCalc.Core/ExpressionEvaluator.cs
```diff
@@ -41,7 +41,8 @@ public int CompareUsingMostPreciseType(object? a, object? b)
     
     protected MathHelperOptions MathHelperOptions => new(Context.CultureInfo,
```
gumbarros, line 42: Perfect, but you need to do the same thing at `AsyncExpressionEvaluator`

gumbarros:
> It seems we need docs versioning for v4 and v5

If I re-add the event handlers at the parameters PR, no hardcore breaking changes and no need for two docs I think

src/NCalc.Core/ExpressionEvaluator.cs
Bykiev: Nice catch, I'll fix it tomorrow evening (I'm on the phone now)

gumbarros: No need, for the change, this is on the base class for both, sorry

gumbarros (approved):

PR #263, opened 2024-06-19
Update EvaluationVisitor.cs

theebow:
Extending "plus" functionnality to manage right string for formula type : [number] + [string]

Usefull in the case of formula evaluator.

### DIFF
--- a/src/NCalc.Async/Visitors/AsyncEvaluationVisitor.cs
+++ b/src/NCalc.Async/Visitors/AsyncEvaluationVisitor.cs
@@ -86,7 +86,7 @@ public class AsyncEvaluationVisitor : ExpressionEvaluator, IAsyncEvaluationVisit
                 break;
 
             case BinaryExpressionType.Plus:
-                if (await leftValue.Value is string)
+                if (await leftValue.Value is string || await rightValue.Value is string)
                 {
                     Result = string.Concat(await leftValue.Value, await rightValue.Value);
                 }
--- a/src/NCalc.Sync/Visitors/EvaluationVisitor.cs
+++ b/src/NCalc.Sync/Visitors/EvaluationVisitor.cs
@@ -92,7 +92,7 @@ public class EvaluationVisitor : ExpressionEvaluator, IEvaluationVisitor
                 break;
 
             case BinaryExpressionType.Plus:
-                if (leftValue.Value is string)
+                if (leftValue.Value is string || rightValue.Value is string)
                 {
                     Result = string.Concat(leftValue.Value, rightValue.Value);
                 }


gumbarros:
Good morning, you also need to update `AsyncEvaluationVisitor`

theebow:
> Hi, can you please add a test case?

`    [Fact]
    public void EvaluateFormulaWithParameters()
    {
        var e = new Expression("[NOORDRE] + [RACINE]", CultureInfo.InvariantCulture);
        e.Parameters["RACINE"] = "X";
        e.Parameters["NOORDRE"] = 1;

        string myResult = e.Evaluate().ToString();
        System.Diagnostics.Debug.Assert(string.Compare("1X", myResult, true) == 0);
    }
`

This sample crash here : 
image

theebow:
> Good morning, you also need to update `AsyncEvaluationVisitor`

i'm not github expert^^ but i'm going to try

theebow:
> > Good morning, you also need to update `AsyncEvaluationVisitor`
> 
> i'm not github expert^^ but i'm going to try

@gumbarros i found how to, it's ok

Bykiev:
This change is breaking. Before this change when the left value is not a string it'll try to convert right string or char parameter to number. On the one side, in C# it's possible to concat numbers and strings, on the other side - it can break the existing apps.

@gumbarros, what do you think?

gumbarros:
> This change is breaking. Before this change when the left value is not a string it'll try to convert right string or char parameter to number. On the one side, in C# it's possible to concat numbers and strings and the result will be concatenated strings, on the other side - it can break the existing apps.
> 
> Upd. `1 + '1'` in C# results to 50, because it's getting a code of char. So, the behaviour of NCalc is different. I vote for stay it as is.
> 
> @gumbarros, what do you think?

Good points, maybe a `ExpressionOptions.CharAsNumber` so we can have both behaviors without breaking? Idk if it's worth.

Bykiev:
But what to do if one of the parameters is number and another one is string? I think it should be `ExpressionOptions.StringAsNumber` and IIRC chars in NCalc always converted to strings. But anyway, it'll introduce a breaking change as we need to convert the right parameter to number if needed

Bykiev:
@gumbarros, sould we introduce a new option or leave it as is, what do you think?

gumbarros:
> @gumbarros, sould we introduce a new option or leave it as is, what do you think?

I still don't understand how it would work, is this the proposed behavior of the new option?

Without `ExpressionOptions.StringAsNumber`:

`1 + "1" = NCalcEvaluationException`

With `ExpressionOptions.StringAsNumber`:

`1 + "1" = 2`

Bykiev:
> > @gumbarros, sould we introduce a new option or leave it as is, what do you think?
> 
> I still don't understand how it would work, is this the proposed behavior of the new option?
> 
> Without `ExpressionOptions.StringAsNumber`:
> 
> `1 + "1" = NCalcEvaluationException`
> 
> With `ExpressionOptions.StringAsNumber`:
> 
> `1 + "1" = 2`

I think we should name it `ExpressionOptions.StringConcat`, it should be working this way

```
if (options.HasFlag(ExpressionOptions.StringConcat))
{
    Result = string.Concat(leftValue.Value, rightValue.Value);
}
else
// try to convert to numbers
```

gumbarros:
> > > @gumbarros, sould we introduce a new option or leave it as is, what do you think?
> > 
> > 
> > I still don't understand how it would work, is this the proposed behavior of the new option?
> > Without `ExpressionOptions.StringAsNumber`:
> > `1 + "1" = NCalcEvaluationException`
> > With `ExpressionOptions.StringAsNumber`:
> > `1 + "1" = 2`
> 
> I think we should name it `ExpressionOptions.StringConcat`, it should be working this way
> 
> ```
> if (options.HasFlag(ExpressionOptions.StringConcat))
> {
>     Result = string.Concat(leftValue.Value, rightValue.Value);
> }
> else
> // try to convert to numbers
> ```

With StringConcat 1+1 would become 11? Or just if one of the values are a string like '1' + 1 = 11?

gumbarros:
> I think 1+1 should be "11" with StringConcat
Yes, what I was thinking, just missed the ". It's not a use case for me but as long this happens only when the option is enabled, I see no problem.

Bykiev:
Of course, it's a breaking change again as "1" + 1 will be 2 (not "11" as now) without this option

gumbarros:
> Of course, it's a breaking change again as "1" + 1 will be 2 (not "11" as now) without this option

I think it's more a fix than a breaking change 😄. But the "canon" CLR evaluation would be this right? 

image

Bykiev:
If I remember correctly in NCalc chars are always treating as strings, if we want a CLR behaviour we need to introduce a bigger breaking changes 😄

gumbarros:
> If I remember correctly in NCalc chars are always treating as strings, if we want a CLR behaviour we need to introduce a bigger breaking changes 😄

Makes sense, because there is no `ValueType.Char`, so I agree with what you proposed.

PR #266, opened 2024-06-27
Added `Id` property to `Identifier`

### LINKED ISSUE #151: Support function effects outside of expression

gumbarros:
This closes #151

### DIFF
--- a/src/NCalc.Sync/ExpressionFunctionData.cs
+++ b/src/NCalc.Sync/ExpressionFunctionData.cs
@@ -1,0 +1,14 @@
+namespace NCalc;
+
+public class ExpressionFunctionData(Guid id, Expression[] arguments, ExpressionContext context)
+{
+    public Guid Id { get; } = id;
+    private Expression[] Arguments { get; } = arguments;
+    public ExpressionContext Context { get; } = context;
+    
+    public Expression this[int index]
+    {
+        get => Arguments[index];
+        set => Arguments[index] = value;
+    }
+}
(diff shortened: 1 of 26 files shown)


gumbarros:
@sklose check the unit tests, this will solve your case right? Thanks for the original implementation

Bykiev:
> @sklose check the unit tests, this will solve your case right? Thanks for the original implementation

did you mean @slobber?

Bykiev (approved):
LGTM

src/NCalc.Sync/ExpressionFunctionData.cs
```diff
@@ -0,0 +1,14 @@
+namespace NCalc;
+
+public class ExpressionFunctionData(Guid id, Expression[] arguments, ExpressionContext context)
```
gmcchessney, line 3: I don't see any way to get the number of arguments, since `Arguments` is private.

Issue #164, opened 2024-05-23
Docs: Write an article about how NCalc works 


gumbarros:
(String->Parlot Parser->EvaluationVisitor->Result)

gumbarros:
Closed via #272

PR #272, opened 2024-06-30
Visitor pattern is now stateless with generics

gumbarros:


gumbarros:
The codebase is so much better, but performance is 2x slower, trying to see how to have this without a performance regression.

Bykiev:
In fact there shouldn't be any performance difference with generics. Looks much better with generics, nice job!

gumbarros:
Added architecture docs. This closes #164

Bykiev (approved):

PR #274, opened 2024-07-01
Small refactoring of Expression.IterateParameters

Bykiev:
Small refactoring of Expression.IterateParameters

### DIFF
--- a/src/NCalc.Async/AsyncExpression.cs
+++ b/src/NCalc.Async/AsyncExpression.cs
@@ -249,9 +249,7 @@ public class AsyncExpression
             {
                 parameterEnumerators.Add(parameter.Key, enumerable.GetEnumerator());
 
-                int localSize = 0;
-                foreach (var o in enumerable)
-                    localSize++;
+                int localSize = enumerable.Cast<object>().Count();
 
                 if (size == null)
                     size = localSize;
--- a/src/NCalc.Async/AsyncExpression.cs
+++ b/src/NCalc.Async/AsyncExpression.cs
@@ -247,9 +247,10 @@ public class AsyncExpression
         {
             if (parameter.Value is IEnumerable enumerable)
             {
+                var list = enumerable as List<object> ?? enumerable.Cast<object>().ToList();
                 parameterEnumerators.Add(parameter.Key, enumerable.GetEnumerator());
 
-                int localSize = enumerable.Cast<object>().Count();
+                int localSize = list.Count;
 
                 if (size == null)
                     size = localSize;
--- a/src/NCalc.Async/AsyncExpression.cs
+++ b/src/NCalc.Async/AsyncExpression.cs
@@ -248,7 +248,7 @@ public class AsyncExpression
             if (parameter.Value is IEnumerable enumerable)
             {
                 var list = enumerable as List<object> ?? enumerable.Cast<object>().ToList();
-                parameterEnumerators.Add(parameter.Key, enumerable.GetEnumerator());
+                parameterEnumerators.Add(parameter.Key, list.GetEnumerator());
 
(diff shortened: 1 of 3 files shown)


src/NCalc.Async/AsyncExpression.cs
```diff
@@ -242,55 +242,41 @@ public bool HasErrors()
...
         {
-            if (parameter is not IEnumerable enumerable)
-                continue;
+            if (parameter.Value is IEnumerable enumerable)
+            {
+                parameterEnumerators.Add(parameter.Key, enumerable.GetEnumerator());
 
-            var localsize = enumerable.Cast<object>().Count();
+                int localSize = 0;
+                foreach (var o in enumerable)
```
gumbarros, line 253: Possible multiple enumeration warning here
https://www.jetbrains.com/help/rider/PossibleMultipleEnumeration.html

gumbarros (changes requested):

src/NCalc.Async/AsyncExpression.cs
Bykiev: Thanks, I'll check it later, need to benchmark casting vs iterating

Bykiev: I did a benchmark with 1000 elements and casting is really faster, I'll rollback my change:

```
| Method      | Mean        | Error      | StdDev     | Rank | Allocated |
|------------ |------------:|-----------:|-----------:|-----:|----------:|
| TestCast    |    15.20 ns |   0.220 ns |   0.253 ns |    1 |         - |
| TestIterate | 8,159.16 ns | 160.834 ns | 225.467 ns |    2 |      40 B |
```

gumbarros: I think just a `ToList` is missing at `var localsize = enumerable.Cast<object>().Count();` to solve the warning

Bykiev: I think it's a false positive, just reverted casting

gumbarros: No, I agree with the warning, it's possible to solve 2 warnings this way:

image
image


image

Bykiev: Agreed, as we're looping through iterator it'll automatically disposed, but still there is a case when the size of arrays are not equal and we still need to dispose enumerator, isn't it?

gumbarros: No, the dispose is because the type of the `IEnumerator` is not known. With a `List` we know it never needs to be disposed.

src/NCalc.Async/AsyncExpression.cs
```diff
@@ -242,55 +242,40 @@ public bool HasErrors()
...
 
-        foreach (var parameter in Parameters.Values)
+        foreach (var parameter in Parameters)
         {
-            if (parameter is not IEnumerable enumerable)
-                continue;
+            if (parameter.Value is IEnumerable enumerable)
+            {
+                var list = enumerable as List<object> ?? enumerable.Cast<object>().ToList();
+                parameterEnumerators.Add(parameter.Key, enumerable.GetEnumerator());
```
gumbarros, line 251: You almost made it:

Wrong:
image
image

gumbarros (changes requested):

gumbarros (approved):

gumbarros:
Just saw I had permission in your repo and commited the warning fix, ok?

gumbarros:
Also please review dd893eccf4f377074acc8d8ce3516ae4d4762b00

Bykiev:
Nice addition, thanks! I think, now it's can be merged

PR #276, opened 2024-07-03
Add new `ExpressionOptions.StringConcat` to concat values as string

Bykiev:
Add new `ExpressionOptions.StringConcat` option to concat values as string. This is a breaking change. In previous versions if the left value is string the result was a string concatenation, but if the left is not string the result will be an arithmetic addition. After this change you should explicitly specify this option for string concatenation.

### DIFF
--- a/docs/articles/string_concatenation.md
+++ b/docs/articles/string_concatenation.md
@@ -1,6 +1,6 @@
 ﻿# String concatenation
 
-If the left or right values of expression are strings and <xref:NCalc.ExpressionOptions.StringConcat> option specified the result will be a string concatenation. Otherwise the arithmetic addition will be performed event with <xref:NCalc.ExpressionOptions.StringConcat> option specified.
+If the <xref:NCalc.ExpressionOptions.StringConcat> option specified the result will be a string concatenation. Otherwise it will try to perform an arithmetic addition if possible or if both values are strings then the result will be a string concatenation.
 
 ```c#
 try
--- a/docs/articles/string_concatenation.md
+++ b/docs/articles/string_concatenation.md
@@ -1,6 +1,6 @@
 ﻿# String concatenation
 
-If the both left and right values of expression are string the result will be a string concatenation. If not, you can still concatenate values as a string by using <xref:NCalc.ExpressionOptions.StringConcat> option.
+If the left or right values of expression are strings and <xref:NCalc.ExpressionOptions.StringConcat> option specified the result will be a string concatenation. Otherwise the arithmetic addition will be performed event with <xref:NCalc.ExpressionOptions.StringConcat> option specified.
 
 ```c#
 try
--- a/docs/articles/string_concatenation.md
+++ b/docs/articles/string_concatenation.md
@@ -1,11 +1,11 @@
 ﻿# String concatenation
 
-To concatenate values as a string you should use <xref:NCalc.ExpressionOptions.StringConcat>.
+If the both left and right values of expression are string the result will be a string concatenation. If not, you can still concatenate values as a string by using <xref:NCalc.ExpressionOptions.StringConcat> option.
 
 ```c#
 try
 {
-   var expression = new Expression($"5 + \"1\"", ExpressionOptions.StringConcat);
+   var expression = new Expression($"5 +'1'", ExpressionOptions.StringConcat);
 }
 catch (OverflowException ex) { }
 ```
--- a/docs/articles/string_concatenation.md
+++ b/docs/articles/string_concatenation.md
@@ -1,0 +1,11 @@
+﻿# String concatenation
+
+To concatenate values as a string you should use <xref:NCalc.ExpressionOptions.StringConcat>.
+
+```c#
+try
+{
+   var expression = new Expression($"5 + \"1\"", ExpressionOptions.StringConcat);
+}
+catch (OverflowException ex) { }
+```
--- a/src/NCalc.Core/ExpressionOptions.cs
+++ b/src/NCalc.Core/ExpressionOptions.cs
@@ -62,5 +62,10 @@ public enum ExpressionOptions
     /// <summary>
     /// Check for arithmetic binary operation overflow
     /// </summary>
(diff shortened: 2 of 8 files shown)


test/NCalc.Tests/AsyncTests.cs
```diff
@@ -33,7 +33,7 @@ public async Task ShouldEvaluateAsyncFunction()
     [Fact]
     public async Task ShouldEvaluateAsyncParameter()
     {
-        var expression = new AsyncExpression("(a + b) == 'Leo'");
+        var expression = new AsyncExpression("(a + b) == 'Leo'", ExpressionOptions.StringConcat);
```
gumbarros, line 36: This is the breaking change right? Expressions like this would need ExpressionOptions.StringConcat?

gumbarros:
To prevent breaking changes and lots of headaches because of this hot path what about:

'L' + 'eo' -> Always concat regardless of `ExpressionOptions.StringConcat`
1 + '2' -> Same behavior as old versions with `ExpressionOptions.StringConcat` off and '12' with `StringConcat on
'1' + '2' -> 3  with `ExpressionOptions.StringConcat` off and '12' with `StringConcat on

Bykiev:
> Or, to prevent a breaking change, invert the behavior creating `ExpressionOptions.StringAddition`

Sorry, didn't get the idea how it will overcome the breaking change...

gumbarros:
> > Or, to prevent a breaking change, invert the behavior creating `ExpressionOptions.StringAddition`
> 
> Sorry, didn't get the idea how it will overcome the breaking change...

I believe that this line of reasoning would not work either. I need some time to think about this PR, I believe it is very critical.

gumbarros:
> 'L' + 'eo' -> Always concat regardless of `StringConcat`
> 1 + '2' -> Same behavior as old versions with `StringConcat` off and '12' with `StringConcat` on
> '1' + '2' -> 3 with `StringConcat` off and '12' with `StringConcat` on

Would this work?

Bykiev:
> > 'L' + 'eo' -> Always concat regardless of `StringConcat`
> > 1 + '2' -> Same behavior as old versions with `StringConcat` off and '12' with `StringConcat` on
> > '1' + '2' -> 3 with `StringConcat` off and '12' with `StringConcat` on
> 
> Would this work?

I think it's possible, will update the PR soon

docs/articles/string_concatenation.md
```diff
@@ -0,0 +1,11 @@
+﻿# String concatenation
+
+If the both left and right values of expression are string the result will be a string concatenation. If not, you can still concatenate values as a string by using <xref:NCalc.ExpressionOptions.StringConcat> option.
+
+```c#
+try
+{
+   var expression = new Expression($"5 +'1'", ExpressionOptions.StringConcat);
+}
+catch (OverflowException ex) { }
```
gumbarros, line 10: Sorry for the small nitpick, but I think this catch is not needed. I would improve this article with the example of the PR
``` 
'L' + 'eo' -> Always concat regardless of StringConcat
1 + '2' -> Same behavior as old versions with StringConcat off and '12' with StringConcat on
'1' + '2' -> 3 with StringConcat off and '12' with StringConcat on
```

src/NCalc.Core/ExpressionOptions.cs
```diff
@@ -62,5 +62,10 @@ public enum ExpressionOptions
     /// <summary>
     /// Check for arithmetic binary operation overflow
     /// </summary>
-    OverflowProtection = 1 << 10
+    OverflowProtection = 1 << 10,
+
+    /// <summary>
+    /// Concat values as strings instead of arithmetic addition
+    /// </summary>
```
gumbarros, line 69: What about `Prioritizes string concatenation over arithmetic addition when one of the values ​​is a string and the other is a number.`?

gumbarros (changes requested):
Nice, I think now we are going to prevent headaches with a runtime breaking change.

src/NCalc.Core/ExpressionOptions.cs
Bykiev: It makes sense, we just need to decide which one is better

gumbarros: > which one is better

Sorry I did not understand, what comment is better or behavior? I think the non breaking behavior is better

Bykiev: Unfortunately, there is no non-breaking changes. I've updated the PR to match this logic:

```
'L' + 'eo' -> Always concat regardless of StringConcat
1 + '2' -> Same behavior as old versions with StringConcat off and '12' with StringConcat on
'1' + '2' -> 3 with StringConcat off and '12' with StringConcat on
```

Bykiev: Ooops, seems first case is not working like in a sample, will update the PR again

Bykiev: Should be ok now

gumbarros: > Unfortunately, there is no non-breaking changes. I've updated the PR to match this logic:
> 
> ```
> 'L' + 'eo' -> Always concat regardless of StringConcat
> 1 + '2' -> Same behavior as old versions with StringConcat off and '12' with StringConcat on
> '1' + '2' -> 3 with StringConcat off and '12' with StringConcat on
> ```

But what is breaking in this logic?

Bykiev: The only breaking case is '1'+2, without ExpressionOptions.StringConcat it will be 3, but in current version the results will be '12'

gumbarros: Sorry for the late response, I was out of town. I think this breaking change makes sense. Let's merge?

Bykiev: No problem, just updated the docs to match current logic. If you think it's ok, we can merge it

Issue #238, opened 2024-06-11
`dotnet test` randomly fails sometimes


gumbarros:
![image](https://github.com/ncalc/ncalc/assets/52143624/56c70176-412b-476d-8b86-61b3737fc1eb)
The tests are all successfully but this weirdly happens sometimes.

PR #279, opened 2024-07-08
Added `ExpressionOptions.AllowCharValues`

gumbarros:


Bykiev (commented):
Should we override the logic for addition? Like CLR, 3 + '3' should be 51

Bykiev:
Maybe a big breaking change would be better? Let's make the behaviour of NCalc to match the CLR:
```
3 + '3' = 51
"3" + 5 = "35"
5 + "3" = 53
```

gumbarros:
> Maybe a big breaking change would be better? Let's make the behaviour of NCalc to match the CLR:
> 
> ```
> 3 + '3' = 51
> "3" + 5 = "35"
> 5 + "3" = "53"
> ```

Agree with the behavior, but it will be non-breaking, because `char` values are parsed only when `ExpressionOptions.AllowCharValues` is specified.

Bykiev (approved):

gumbarros:
Something is still wrong. Sometimes the test pass and sometimes it fails, including in my machine.

Bykiev:
> Something is still wrong. Sometimes the test pass and sometimes it fails, including in my machine.

Yes, seems smth is wrong, the merged PR also failed on CI

Bykiev (approved):

gumbarros:
> > Something is still wrong. Sometimes the test pass and sometimes it fails, including in my machine.
> 
> Yes, seems smth is wrong, the merged PR also failed on CI

The `main` error it's just #238, everything is passed. But in this PR, `[InlineData("3 + '3'", ExpressionOptions.AllowCharValues,54)]` does not work in parallel executions but work at single executions.

Bykiev:
As this is not related to this PR it should be addressed in another one, This PR can be merged

gumbarros:
@Bykiev after some rest solved the error. It's because the expression `3 + '3'` is used two times at the unit tests and is cached. `ExpressionOptions.NoCache` solved the problem.

PR #283, opened 2024-07-09
Small performance improvement for `BinaryExpression`

gumbarros:
| Method           | Mean     | Error    | StdDev   | Rank | Gen0   | Allocated |
|----------------- |---------:|---------:|---------:|-----:|-------:|----------:|
| UnsafeExpression | 14.72 us | 0.042 us | 0.039 us |    1 | 0.3967 |   3.24 KB |
| SafeExpression   | 15.12 us | 0.118 us | 0.104 us |    2 | 0.4272 |   3.55 KB |

I think we don't need thread safety as this is a variable and not a property right?


Bykiev (approved):
LGTM

PR #290, opened 2024-07-13
Added `FunctionExtractionVisitor`

gumbarros:


src/NCalc.Core/Visitors/FunctionExtractionVisitor.cs
```diff
@@ -0,0 +1,63 @@
...
+{
+    public List<string> Visit(Identifier identifier) => [];
+
+    public List<string> Visit(LogicalExpressionList list)
+    {
+        var functions = new List<string>();
+        foreach (var value in list)
+        {
+            if (value is not Function function) 
+                continue;
```
Bykiev, line 18: Do we really need `is not Function` with `continue` here? Personally, I don't like `continue` as it makes code hard to read. What do you think?

Bykiev (approved):

src/NCalc.Core/Visitors/FunctionExtractionVisitor.cs
gumbarros: [Inverting the `if` increase the nesting:](https://www.jetbrains.com/help/rider/InvertIf.html)
image

But in this case as the code is small, does not make much difference.

PR #292, opened 2024-07-15
Added `EvaluationHelper` and `in` operator should respect string comparer

gumbarros:
The string comparer part is very important for me. `EvaluationHelper` reduces code duplication of sync and async.


Bykiev (approved):
LGTM

PR #296, opened 2024-07-22
Refactor BuiltInFunctionHelper

Bykiev:
Some refactoring of `BuiltInFunctionHelper` to remove code duplication


gumbarros:
Sorry, I'm not a fan of this PR for 2 reasons:

1. Confusing public API: It's possible to confuse consumers with the additional Evaluate methods on both `async` and `sync`
image

2. Possible deadlock
image

I used one time the `GetAwaiter().GetResult()` strategy at a legacy WebForms app and the request literally died waiting forever at this method call. 

I agree with the intent of the PR and I'm also studying some way to remove this huge code duplication, but IDK if it's possible without some drawback.

gumbarros:
One possible solution for the whole `async/sync` would be a source generator, but I never worked with one.

Bykiev:
Agreed, it's not perfect, but seems the only solution without code duplication. I'll try to research another ways

Bykiev:
@gumbarros, to solve the issue with confusing API, we can make a common `Evaluate()` method, this will be a breaking change. What do you think?

gumbarros:
> @gumbarros, to solve the issue with confusing API, we can make a common `Evaluate()` method, this will be a breaking change. What do you think?

Hmm Idk, it will force and `async` on `sync`? I think this is impossible to solve =\

PR #321, opened 2024-08-24
Add "no string type coercion" option

alexwarren:
This PR adds a new option `NoStringTypeCoercion`.

Currently by default NCalc will try to interpret strings like `"1"` as numbers when performing addition. This leads to some potentially unwanted behaviour. The option `StringConcat` was added to bypass this option by _always_ performing string concatenation, but this is also not always desirable.

With this new option enabled, strings are always interpreted as strings, which means with the `+` operator, string concatenation will _only_ be performed if _either_ value is a string.

### DIFF
--- a/src/NCalc.Core/ExpressionOptions.cs
+++ b/src/NCalc.Core/ExpressionOptions.cs
@@ -69,5 +69,10 @@ public enum ExpressionOptions
     /// <summary>
     /// Parse single quoted strings as <see cref="char"/>
     /// </summary>
-    AllowCharValues = 1 << 12
+    AllowCharValues = 1 << 12,
+
+    /// <summary>
+    /// Disables coercion of string values to other types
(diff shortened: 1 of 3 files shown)


gumbarros (approved):
LGTM, thanks for the contribution.

Bykiev (approved):
I see it's very specific usage, but as there is no any breaking changes I see no problems to add new option

gumbarros:
> I see it's very specific usage, but as there is no any breaking changes I see no problems to add new option

Probably this was broken at v5.1 with the `StringConcat` option and new `NoStringTypeCoercion` restores the old v5 behavior.

Bykiev:
Yes, we've change the behaviour and only if both of the values can be converted to number is allowed. Feel free to merge this PR (as the new option is added we should increase the minor version number)

PR #328, opened 2024-08-31
Use `Microsoft.Extensions.Logging` instead of hard-coded logging

### LINKED ISSUE #267: Performance issue resurfaced
**Describe the bug**
Exactly same issue as described here: https://github.com/ncalc/ncalc/issues/51
The previous fix seems to be missing from the Core and Async variant libraries resulting in the same bug as described in the previous issue.
Resurfaced because of this change: https://github.com/ncalc/ncalc/pull/207

- Affects NCalc version 4.3.0 to 4.3.3
The bug is not present in version 4.2.1

Running my application is miserable now

I shouldn't have to see NCalc Trace output when debugging my application or in production
image

The fix is probably just adding `<DefineConstants>$(DefineConstants.Replace("TRACE", "NOTRACE"))</DefineConstants>` to the .csproj files

gumbarros:
This closes #267

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -35,7 +35,6 @@
   </ItemGroup>
   <PropertyGroup Condition="'$(Configuration)'=='Release' Or '$(Configuration)'=='SignedRelease'">
     <Deterministic>true</Deterministic>
-    <DefineConstants>$(DefineConstants.Replace("TRACE", "NOTRACE"))</DefineConstants>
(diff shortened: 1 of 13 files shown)


Bykiev (approved):
Good job, thanks!

eero-dev:
Thank you, could you make a release so we can try it out?

gumbarros:
> Thank you, could you make a release so we can try it out?

Yes, I have some tasks to do at my job, until the end of the day I will create a release.

PR #329, opened 2024-09-02
`ExpressionOptions.NoStringTypeCoercion` should respect `in` operator

gumbarros:


gumbarros:
Needing this at prod for @JJConsulting

PR #330, opened 2024-09-03
Fix `in` operator when there is an empty string

gumbarros:


gumbarros:
The reason of the bug is this:

image

Expression String
"6 in ('')"

becomes at CLR

"6".Contains("")

and this is true for the CLR, but makes no sense as an int.

PR #341, opened 2024-09-30
Improve number parser readability

gumbarros:


Bykiev:
LGTM, but I'm still confused: should we throw an overflow exception instead of returning a double value when `DecimalAsDefault` is used?

src/NCalc.Core/Parser/LogicalExpressionParser.cs
```diff
@@ -479,7 +407,78 @@ static LogicalExpressionParser()
...
+            else
+                result = (double)res;
+        }
+
+        if (decimalPart != null || exponentPart != null)
+        {
+            return new ValueExpression(result);
+        }
+
+        return new ValueExpression((long)result);
```
Bykiev, line 478: Missed this in previous review, can you please just return `integralValue` here?

gumbarros: Good catch

src/NCalc.Core/Parser/LogicalExpressionParser.cs
```diff
@@ -479,7 +407,78 @@ static LogicalExpressionParser()
...
+            // decimal part?
+            if (decimalPart != null && decimalPart.Value != 0)
+            {
+                var digits = Math.Floor(Math.Log10(decimalPart.Value) + 1) + zeroCount;
+                result1 += decimalPart.Value / (decimal)Math.Pow(10, digits);
+            }
+
+            // exponent part?
+            if (exponentPart == null) return new ValueExpression(result1);
+            var left = BigDecimal.Parse((double)result1);
```
Bykiev, line 430: I did a bench, and this seems to be 2x slower than `var left = new BigDecimal(result1);`

| Method    | Mean     | Error     | StdDev    | Ratio | RatioSD | Gen0   | Allocated | Alloc Ratio |
|---------- |---------:|----------:|----------:|------:|--------:|-------:|----------:|------------:|
| TestNew   | 1.390 us | 0.0166 us | 0.0227 us |  1.00 |    0.02 | 0.3395 |     536 B |        1.00 |
| TestParse | 2.339 us | 0.0126 us | 0.0105 us |  1.68 |    0.03 | 0.7133 |    1120 B |        2.09 |

gumbarros:
> LGTM, but I'm still confused: should we throw an overflow exception instead of returning a double value on overflow when `DecimalAsDefault` is used?

Hmm, maybe handle this with `ExpressionOptions.OverflowProtection`?

Bykiev:
> > LGTM, but I'm still confused: should we throw an overflow exception instead of returning a double value on overflow when `DecimalAsDefault` is used?
> 
> Hmm, maybe handle this with `ExpressionOptions.OverflowProtection`?

I don't think it's a good idea as it was intended for arithmetic overflow and decimals doesn't support checked maths

gumbarros:
> > > LGTM, but I'm still confused: should we throw an overflow exception instead of returning a double value on overflow when `DecimalAsDefault` is used?
> > 
> > 
> > Hmm, maybe handle this with `ExpressionOptions.OverflowProtection`?
> 
> I don't think it's a good idea as it was intended for arithmetic overflow and decimals doesn't support checked maths

Let's track this in another issue/PR.

Bykiev (approved):

PR #345, opened 2024-10-09
Code Quality: Use .slnx format instead of .sln

gumbarros:
image

Rider officially now supports, so I see no problem :)


gumbarros:
Waiting for CLI support

PR #350, opened 2024-10-25
Fix comparing empty strings with numbers

### LINKED ISSUE #349: Comparing empty strings with int values
_Originally posted at Discord Server by @C7H8_

Hello!

Could someone help me.

I get the NCalc.Expression: The input string was not in a correct format. error message.

Because I'm trying to compare empty string to string.

In C# terms, it would be string.Equals("", "01");

How can I use NCalc to achieve this requirement?

What does the Expression string look like? 

I tried the following writing Expression string, not working.
```
"'' == 01"
"' ' == 01"
"==01"
"" " == 01"
```

Bykiev:
This PR fixes comparing empty strings with int values.
Closes #349

### DIFF
--- a/src/NCalc.Core/Helpers/TypeHelper.cs
+++ b/src/NCalc.Core/Helpers/TypeHelper.cs
@@ -126,12 +126,21 @@ public static class TypeHelper
 
     public static int CompareUsingMostPreciseType(object? a, object? b, ComparisonOptions options)
     {
+        var comparer = GetStringComparer(options);
+
+        if (a is string || b is string)
+        {
+            string? strA = Convert.ToString(a, options.CultureInfo);
+            string? strB = Convert.ToString(b, options.CultureInfo);
+
+            if (string.IsNullOrWhiteSpace(strA) || string.IsNullOrWhiteSpace(strB))
+                return comparer.Compare(strA, strB);
+        }
+
         var mpt = GetMostPreciseType(a?.GetType(), b?.GetType());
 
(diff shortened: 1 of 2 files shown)


gumbarros:
I did something similar, but it will work if you input

"Hello there" == 1

Bykiev:
You're right, we should convert them to string if any of the values is string

Bykiev:
> I did something similar, but it will work if you input
> 
> "Hello there" == 1
> 
> ?

Thanks, I did some tests, and It'll be not possible like in .NET. Finally, I think we shouldn't compare strings with numbers and this PR shouldn't be merged,

gumbarros:
> > I did something similar, but it will work if you input
> > "Hello there" == 1
> > ?
> 
> Thanks, I did some tests, and It'll be not possible like in .NET. Finally, I think we shouldn't compare strings with numbers and this PR shouldn't be merged,

I agree, I was developing something like:

```cs
    /// <summary>
    /// Compares values without performing any type casting, ensuring strict type safety during comparisons.
    /// </summary>
    CompareWithoutCasting = 1 << 15


        protected int Compare(object? a, object? b)
    {
        if (!context.Options.HasFlag(ExpressionOptions.CompareWithoutCasting))
            return CompareUsingMostPreciseType(a, b, context);

        if (context.Options.HasFlag(ExpressionOptions.CaseInsensitiveStringComparer))
            return Comparer.DefaultInvariant.Compare(a,b);

        return Comparer.Default.Compare(a,b);
    }
```
But still getting errors with type conversions

PR #351, opened 2024-10-25
Added `ExpressionOptions.StrictTypeMatching`

### LINKED ISSUE #349: Comparing empty strings with int values

gumbarros:
This closes #349

### DIFF
--- a/src/NCalc.Async/Visitors/AsyncEvaluationVisitor.cs
+++ b/src/NCalc.Async/Visitors/AsyncEvaluationVisitor.cs
@@ -1,12 +1,11 @@
-﻿using System.Numerics;
-using System.Threading;
-using ExtendedNumerics;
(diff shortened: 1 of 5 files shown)


Bykiev (approved):
LGTM

gumbarros:
> LGTM

Thanks for the other PR, the idea wouldn't be possible with it.

PR #363, opened 2024-11-15
Update NCalc logo

gumbarros:
Tried creating a new logo using [.NET guidelines](https://github.com/dotnet/brand/blob/main/dotnet-styleGuide-2024.pdf). What do you think?


Bykiev (approved):
The new logo looks fresh and modern, let's use it!

PR #366, opened 2024-11-25
Update packages

Bykiev:
Updated `Parlot` and `PolySharp` packages

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -13,7 +13,7 @@
       <Version>3.6.146</Version>
       <PrivateAssets>all</PrivateAssets>
     </PackageReference>
-    <PackageReference Include="PolySharp" Version="1.14.1">
+    <PackageReference Include="PolySharp" Version="1.15.0">
       <PrivateAssets>all</PrivateAssets>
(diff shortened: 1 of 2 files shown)


Bykiev:
According to Parlot's changelog the performance should be improved, but the benchmarks shows degraded performance.
@sebastienros, can you please check what's wrong?

sebastienros:
Haven't had time to check for perf so far, but still intend to. Also wondering if ncalc could use the new configuration knobs for parsing numbers... or maybe it's not possible because of BingInt? That could be extensible because I made it so if the type implements the new `INumber`.

sebastienros:
Identified the bug in Parlot that makes the tests failed. But found another one that is breaking YesSql so I will release a new version when both are fixed.

sebastienros:
@gumbarros I think it's also faster (I compared the Benchmarks actions with the previous build).

sebastienros:
Another side effect, is that the compiled mode (for Parlot), when enabled, would take a long time to build the parser, like 10 seconds. I identified the issue in Parlot and fixed it with this version. I ran the tests and all passed. Maybe it would be useful to run the tests with compilation too, at least in PRs by setting a flag.

PR #385, opened 2025-01-31
Update parameters.md

gmkado:
Examples used xor instead of exponential syntax


gumbarros:
Sorry, I did not saw your PR, thanks for the contribution.

Issue #66, opened 2023-03-09
Hexadecimal support


czxinc:
Do Ncalc supports hexadecimal number?

I try this:
`New Expression("0x10020").Evaluate()`.

It throws exception.

petli:
Sorry, there's currently no support for hexadecimal.

PR #388, opened 2025-02-05
Add binary, octal and hexadecimal number support.

slobber:
You can use binary, octal, hexadecimal number in expression, such as `0xad`, `0o17`, `0b0110`.

LIMIT: the value will be translate to `long`, so you cannot set the value large then `long.MaxValue`.


slobber:
fixed issue #66

gumbarros:
Thank you

Issue #391, opened 2025-02-12
CultureInfo does not affect conversion to string


ridago:
**Describe the bug**
Setting CultureInfo for the context does not affect how values are converted to string. In the example below I'd expect NCalc to use the provided culture info when converting to string, instead it uses the global culture info.

**Example Expression**
```C#
var localCulture = (CultureInfo)CultureInfo.InvariantCulture.Clone();
localCulture.NumberFormat.NumberDecimalSeparator = ",";
Thread.CurrentThread.CurrentCulture = localCulture;
var context = new ExpressionContext(ExpressionOptions.StringConcat, CultureInfo.InvariantCulture);
var expression = new Expression("45.6 + [a]", context);
expression.Parameters.Add("a", -12.3);

var result = expression.Evaluate();

// Expected output: "45.6-12.3"
// Actual output: "45,6-12,3"
```

PR #392, opened 2025-02-12
Use CultureInfo in string concatenation

ridago:
This fixes the specific problem outlined in #391, but I don't know if there are more cases where culture info might be ignored.


gumbarros:
Thanks for the contribution

Issue #375, opened 2024-12-15
Help with creating Costume visitor


Beckdotan:
### After my request for an option to change the logic of basic operations in Ncalc was denied, I was trying to update a visitor I had when I want trying to update Ncalc from version 4 to the new version 5.3 I got many issues with my own visitor. 

This is the visitor I currently have that works on version 4

using NCalc;
using NCalc.Domain;
using System;
using System.Collections.Generic;
using System.Globalization;
using System.Linq;

namespace eCountV12.Classes
{

  //following insturctions from here: https://github.com/ncalc/ncalc/issues/72#issuecomment-1794500959
  // logic of expressions can be tested here: https://dotnetfiddle.net/airYJR

  public class EcoosCalcVisitor : EvaluationVisitor
  {
    public EcoosCalcVisitor(EvaluateOptions options = EvaluateOptions.IgnoreCase, CultureInfo cultureInfo = null) :
      base(options, cultureInfo ?? CultureInfo.CurrentCulture)
    {
      CultureInfo = cultureInfo;
      Parameters = new Dictionary<string, object>();
    }

    public Dictionary<string, object> Parameters { get; internal set; }
    private CultureInfo CultureInfo { get; set; }

    /// <summary>
    /// Converts the specified object to a boolean value.
    /// </summary>
    /// <param name="obj">The object to convert to a boolean value.</param>
    /// <returns> Returns `false` if the object is null , or empty string. 
    /// Returns `true` if the object is `true`, all numbers (including 0), or a non-empty string.</returns>
    private static bool ToBoolean(object obj)
    {
      double numericValue;
      //if its numeric value and its 0 return true
      if (double.TryParse(Convert.ToString(obj), out numericValue) && numericValue == 0)
      {
        return true;

      }
      else // follow the ToBoolean defualt values as below:
      {
        return Convert.ToBoolean(obj, null);
        /***
         * Convert.ToBoolean(obj, null):

         * Returns true for:
         * Numeric values (excluding zero): Any non-zero numeric value will be converted to true. For instance, Convert.ToBoolean(5, null) will return true.
         * String values: "true", "True", "TRUE" will be converted to true. Any non-empty string other than those will also be converted to true. For instance, Convert.ToBoolean("true", null) or Convert.ToBoolean("Hello", null) will return true.
         * Non-null objects: Any non-null object reference will be converted to true.

         * Returns false for:
[...]

PR #395, opened 2025-02-13
Added `IEvaluationVisitorFactory` for more granular control over evaluation.

### LINKED ISSUE #394: What is the recommended way to extend evaluation functionality?
### Discussed in https://github.com/ncalc/ncalc/discussions/384

<div type='discussions-op-text'>

<sup>Originally posted by **gmkado** January 30, 2025</sup>
For example, say I have a type `Foo` that supports the plus operator overload.  

I'd like to be able to support expressions like `foo1 + foo2`, but to do that it looks like I need to:

1. Create my own `EvaluationVisitor` which is identical except it adds foo handling in `BinaryExpressionType.Plus`
2. Implement my own `IEvaluationService` so I can pass in my visitor from step 1 into the expression during Evaluate

Alternatively, I could add a function `AddFoo` to the context and users would need to write `AddFoo(foo1, foo2)` instead.

Ideally, we'd be able to extend the existing `EvaluationVisitor`, but that doesn't seem to be possible atm.  Does this sound right?

EDIT: 

I RTFM and found this https://ncalc.github.io/ncalc/articles/architecture.html#evaluation
So I think my interpretation is correct.

I feel like the library could use some more extension points so we're not duplicating entire classes.  I get if you're trying to do something entirely different with the expression (like the serialization visitor mentioned), that it makes sense to create a new visitor.   

I guess I could still use `EvaluationVisitor` but wrap the visiting in a try-catch and then handle Foo types if the evaluation fails.  But then we're catching exceptions which impacts performance.</div>

gumbarros:
This closes #394 #375.

@Beckdotan @gmkado will this solve the problem? Check modified docs and unit tests.


Bykiev:
Hi, good job! The changes are looking good to me, but tests are failing for some reason
`System.IO.FileNotFoundException: Could not find 'mono' host. Make sure that 'mono' is installed on the machine and is available in PATH environment variable.`

Bykiev (approved):

gmkado:
@gumbarros thanks for doing this so quickly!  Yeah I believe it does address my issue

petero-dk:
Fyi I fixed that by extending the factory like this:

```csharp
    public class ObjectEvaluationFactory(
            ILogicalExpressionFactory logicalExpressionFactory,
            ILogicalExpressionCache cache) : IEvaluationVisitorFactory
    {
        public ILogicalExpressionCache Cache => cache;
        public ILogicalExpressionFactory LogicalExpressionFactory => logicalExpressionFactory;
        public EvaluationVisitor Create(ExpressionContext context)
        {
            return new ObjectVisitor(this, context);
        }
    }
```

and then adding a Visit for function expressions like this

```csharp

        public override object Visit(Function function)
        {
            var argsCount = function.Parameters.Count;
            var args = new Expression[argsCount];

            // Don't call parameters right now, instead let the function do it as needed.
            // Some parameters shouldn't be called, for instance, in a if(), the "not" value might be a division by zero
            // Evaluating every value could produce unexpected behaviour
            for (var i = 0; i < argsCount; i++)
            {
                args[i] = new Expression(function.Parameters[i], context, factory.LogicalExpressionFactory, factory.Cache, factory);
            }

            var functionName = function.Identifier.Name;
            var functionArgs = new FunctionArgs(function.Identifier.Id, args);

            OnEvaluateFunction(functionName, functionArgs);

            if (functionArgs.HasResult)
                return functionArgs.Result;

            if (context.Functions.TryGetValue(functionName, out var expressionFunction))
            {
                return expressionFunction(new(function.Identifier.Id, args, context));
            }

            return BuiltInFunctionHelper.Evaluate(functionName, args, context);
        }

```

PR #407, opened 2025-04-23
Improve lambda compilation performance

Bykiev:
This PR improves lambda compilation performance by using FastExpressionCompiler and some code optimizations

// * Summary *

BenchmarkDotNet v0.14.0, Ubuntu 24.04.2 LTS (Noble Numbat)
.NET SDK 9.0.203
  [Host]     : .NET 8.0.15 (8.0.1525.16413), X64 RyuJIT AVX2
  DefaultJob : .NET 8.0.15 (8.0.1525.16413), X64 RyuJIT AVX2

Before:

| Method                   | Mean           | Error       | StdDev      | Rank | Gen0   | Gen1   | Allocated |
|------------------------- |---------------:|------------:|------------:|-----:|-------:|-------:|----------:|
| LambdaWithoutCompilation |      0.3055 ns |   0.0097 ns |   0.0091 ns |    1 |      - |      - |         - |
| Evaluate                 |    737.6502 ns |   8.2438 ns |   7.7113 ns |    2 | 0.1059 |      - |    1776 B |
| LambdaWithCompilation    | 26,107.7541 ns | 296.6398 ns | 262.9635 ns |    3 | 0.2441 | 0.1831 |    5012 B |

After:
| Method                   | Mean        | Error      | StdDev     | Rank | Gen0   | Allocated |
|------------------------- |------------:|-----------:|-----------:|-----:|-------:|----------:|
| LambdaWithoutCompilation |   0.3731 ns |  0.0086 ns |  0.0081 ns |    1 |      - |         - |
| LambdaWithCompilation    | 717.6883 ns |  4.9249 ns |  4.6068 ns |    2 | 0.0248 |     424 B |
| Evaluate                 | 772.2176 ns | 12.6205 ns | 11.8052 ns |    3 | 0.1059 |    1776 B |


gumbarros:
Wow, so it's faster to use `LambdaWithoutCompilation` than a `Evaluate()`?

Bykiev:
If you need to make a lot of evaluations of the same expression with different parameters lambda compilation will be always faster, but the compilation time is quite big, that's why you should cache compilation result.

Bykiev:
So, with FastExpressionCompiler the compilation performance is better, but the evaluation time is bigger (0.4965 ns), but it's very strange, maybe we should use `FastExpressionCompiler.LightExpression.Expression`, but conditional switch will be harder and it'll be easier to implement `LambdaFastExpressionVisitor`, not sure about the performance, but benchmarks [here](https://github.com/dadhi/FastExpressionCompiler) looking very promising

Bykiev:
With `LightExpression` the results are pretty similar, but with better memory allocation:

| Method                   | Mean           | Error       | StdDev      | Rank | Gen0   | Allocated |
|------------------------- |---------------:|------------:|------------:|-----:|-------:|----------:|
| LambdaWithoutCompilation |      0.4928 ns |   0.0136 ns |   0.0128 ns |    1 |      - |         - |
| Evaluate                 |    700.6630 ns |   4.8242 ns |   4.5125 ns |    2 | 0.1059 |    1776 B |
| LambdaWithCompilation    | 17,130.0961 ns | 172.7045 ns | 161.5479 ns |    3 | 0.0610 |    1568 B |

It seems for NCalc the evaluation performance is more important than compilation. 
@gumbarros, what do you think?

Bykiev:
@dadhi, can you please take a look, why does execution is slowly with `CompileFast()` by approx 60%?

dadhi:
@Bykiev I will look... and will start from what are you actually measuring.

dadhi:
@Bykiev To speedup things could you point me to the benchmark code?

dadhi:
@Bykiev I am not sure what is benchmark benchmarking. 
In the  `LambdaWithCompilation`  it is Creates Expression + Compiles + + Invokes it.  Yes, it will be slow if you are doing it each time. 
Maybe you supposed to benchmark just the Invoke for cached compiled delegate?

Btw,
I've looked into the code and think there are a lot of optimizations can be done to make Create+Compile faster.

Bykiev:
> @Bykiev I am not sure what is benchmark benchmarking.
> 
> In the `LambdaWithCompilation` it Creates Expression + Compiles + Invokes it. Yes, it will be slow if you are doing it each time. Maybe you supposed to benchmark just the Invoke for cached compiled delegate?

It's ok, you should check [`LambdaWithoutCompilation`](https://github.com/ncalc/ncalc/blob/master/test/NCalc.Benchmarks/EvaluateVsLambdaBenchmark.cs) benchmark. Now it's slower than it was with compile.


> Btw,
> I've looked into the code and think there are a lot of optimizations can be done to make Create+Compile faster.

We're open for your ideas :)

Bykiev:
@dadhi, (just for notification, not sure if you're seeing notifications from this thread)

dadhi:
So, it is `Compile` vs `CompileFast`, right?

>  (just for notification, not sure if you're seeing notifications from this thread)
Yes, I see the notifications without mention.

Bykiev:
> So, it is Compile vs CompileFast, right?

Yes, the compilation performance is really better with `CompileFast()` (memory allocation is 3 times lower and overall time is better by 30%), but invoking (`LambdaWithoutCompilation` benchmark) is slower

dadhi:
@Bykiev I have released FEC v5.2.0 with the improvements.
My latest benchmarks can be found here: https://github.com/dadhi/FastExpressionCompiler/issues/472
Specifically, look at 2 last benchmarks in the description for the compilation and invocation.

Bykiev:
Thank you, I've updated my initial message with benchmarks and now results are very impressive!

src/NCalc.Sync/Expression.Lambda.cs
```diff
@@ -7,6 +9,13 @@ namespace NCalc;
 
 public partial class Expression
 {
+    internal static readonly bool UseSystemLynqCompiler;
```
gumbarros, line 12: I _think_ the correct name should be `UseSystemLinqCompiler`

gumbarros (changes requested):

gumbarros (approved):

Bykiev:
@sebastienros, @sklose, @yallie, whould you like to review the PR?

sklose:
looks like a nice improvement to me, kudos @Bykiev 🚀

sklose (approved):

PR #421, opened 2025-05-19
Add integer parser

Bykiev:
Integer parser should improve performance for integer parsing


gumbarros:
Incrible job these days, I will try to publish a version soon

PR #428, opened 2025-06-03
Add multiple features as listed in description

emayevski:
I am offering several enhancements, both in parsing and operations, namely:

* Advanced date and time parsing, which takes into account culture settings (current or specific culture or custom separators) and supports times with or without seconds as well as 12-hour time
* Currency support, which takes into account culture settings (current or specific culture or custom symbols). The idea is that the parser should drop the currency symbol when the expression is copied by the user from some source and includes this currency symbol (don't force people remove manually what the parser can remove automatically)
* Underscores in numbers and currency . Modern programming languages support underscores for readability. Support in binary, octal, and hex numbers is built-in, while support in decimals requires a patch in Parlot. 
* Custom decimal and group separators in numbers and currency. Requires more attention to optionally support both comma and dot as separators, but this also requires modifying Parlot parser or using the alternative parsing for decimal and double numbers.
* C-Style octal literals.
* Result Reference character. A pseudo-function that would let a user application return some value, such as a result of the previous calculation. This is handy when an expression should include this result multiple times.  
* Percent calculations.  A new PercentExpression was introduced and enhancements were made in the handling of math expressions. 

Other notable changes include 
* the possibility to re-initialize the parser before parsing an expression (the previous implementation used a fixed parser created by the static constructor);
*  new flags in ExpressionOptions to skip date and GUID parsers in order to speed up parsing (as per Andrey Bykiev's suggestion). More similar flags can be added separately. 

I tried to move this code to a plugin, but unfortunately, extending visitors with support of percent calculations is not possible. Also, many classes are sealed or internal, making them impossible to use or descend from in a plugin project. So, the only way forward for these enhancements can be to go to the main library.


Bykiev:
@emayevski, thank you for you inputs, can you please resolve conflicts with the main branch?

emayevski:
> @emayevski, thank you for your inputs, can you please resolve conflicts with the main branch?

@Bykiev This can be tricky - I would need to either override all your changes related to cultures in dates (when I started, your pull request was not even submitted) or somehow incorporate them into the pull request.  I don't see a good way to do this now; got to think more about the possible ways of doing this in the evening. Maybe, I need to delete this pull request, rework the code on my side, and create a new pull request. 

@gumbarros Could you please review my request as is and tell me whether you would accept it when the conflicts are resolved? If you don't accept it, it's not worth spending time on the conflict resolution. If you'd like to discuss certain functionality, let's do this.

Bykiev:
By the way, can you please take a look at [PanoramicData.NCalcExtensions](https://github.com/panoramicdata/PanoramicData.NCalcExtensions) project? I love their idea of implementing additional features via functions, that's why I want to make such a breaking change and remove guid, dates from parser (in NCalc v6). Seems percent  calculations can't be done this way and can be a part of NCalc.

emayevski:
> By the way, can you please take a look at [PanoramicData.NCalcExtensions](https://github.com/panoramicdata/PanoramicData.NCalcExtensions) project? I love their idea of implementing additional features via functions, that's why I want to make such a breaking change and remove guid, dates from parser (in NCalc v6). Seems percent calculations can't be done this way and can be a part of NCalc.

I have their project checked out. Functions are easier to add. Extending the parser is more complicated and requires duplicating plenty of code due to current code restrictions (sealed and internal classes, etc.), and percent calculations require a modification of NCalc itself.  This is why I switched from the idea of creating an advanced plugin (which I started and put on hold) to this pull request.

gumbarros:
> > @emayevski, thank you for your inputs, can you please resolve conflicts with the main branch?
> 
> @Bykiev This can be tricky - I would need to either override all your changes related to cultures in dates (when I started, your pull request was not even submitted) or somehow incorporate them into the pull request. I don't see a good way to do this now; got to think more about the possible ways of doing this in the evening. Maybe, I need to delete this pull request, rework the code on my side, and create a new pull request.
> 
> @gumbarros Could you please review my request as is and tell me whether you would accept it when the conflicts are resolved? If you don't accept it, it's not worth spending time on the conflict resolution. If you'd like to discuss certain functionality, let's do this.

Wow, congratulations for the features 🚀, I will sure merge it after the conflicts

gumbarros:
I'm fully on board with having advanced features. I think it's great to have them available to solve complex problems. However, I wonder why `AdvExpressionOptions` isn't part of `ExpressionContext`. Including it there might lead to a simpler and more intuitive public API.

emayevski:
I've made the final fixes to make all tests pass when run all together.

emayevski:
@Bykiev Was it you who ran the tests (this is what Github states)? Let me ask, how I can reproduce the problem: locally, all tests pass just fine, both from the console and in Test Explorer. I just don't know what to fix. 

image

Bykiev:
Yes, I'm running workflow manually, but can't find what's wrong...

Bykiev:
I've created https://github.com/ncalc/ncalc/pull/432 to fix tests output

emayevski:
Closing this pull request in favor of #435 , which has more cool new features.

PR #435, opened 2025-06-12
Add multiple features as listed in description

emayevski:
I am offering several enhancements, both in parsing and operations, namely:

* Assignable parameters including shortcut operators (+=, etc.)
* Statement sequences (useful together with parameter assignments)
* Advanced date and time parsing, which takes into account culture settings (current or specific culture or custom separators) and supports times with or without seconds as well as 12-hour time
* Parsing of humane period expressions like "3 weeks 2 days 5 hours" (period identifiers are customizable and multiple identifier per period are supported)
* Basic calculations with dates and time spans - one can add and subtract dates and times. Without these operations, date and time values are of little use (if only with custom functions).
* Currency support, which takes into account culture settings (current or specific culture or custom symbols). The idea is that the parser should drop the currency symbol when the expression is copied by the user from some source and includes this currency symbol (don't force people remove manually what the parser can remove automatically)
* Underscores in numbers and currency . Modern programming languages support underscores for readability. Support in binary, octal, and hex numbers is built-in, while support in decimals requires a patch in Parlot.
* Custom decimal and group separators in numbers and currency. Requires more attention to optionally support both comma and dot as separators, but this also requires modifying Parlot parser or using the alternative parsing for decimal and double numbers.
* C-Style octal literals.
* Result Reference character. A pseudo-function that would let a user application return some value, such as a result of the previous calculation. This is handy when an expression should include this result multiple times.
* Percent calculations. A new PercentExpression was introduced and enhancements were made in the handling of math expressions.
* Factorials (both regular and any complex factorials on integers are supported).
* Logical XOR operations
* Certain Unicode characters can be used as operators
* Lowercase lookup for parameter and function names

Other notable changes include

* the possibility to re-initialize the parser before parsing an expression (the previous implementation used a fixed parser created by the static constructor);
* new flags in ExpressionOptions to skip date and GUID parsers in order to speed up parsing (as per Andrey Bykiev's suggestion). More similar flags can be added separately.

All these enhancements are supported in synchronous, asynchronous, and lambda modes, but only with a Parlot-based parser.


gumbarros:
Wow, thank you very much for the contributions, it will have things helpful for many users. Can you fix the conflicts @emayevski? And sorry for the late response, lots of personal and professoinal problems.

src/NCalc.Core/Domain/Percent.cs
```diff
@@ -0,0 +1,42 @@
+﻿using NCalc.Exceptions;
+using NCalc.Visitors;
+
+namespace NCalc.Domain
+{
```
gumbarros, line 5: Small nitpick, can you use file scoped namespace here?

gumbarros (changes requested):

src/NCalc.Core/Domain/Percent.cs
emayevski: Done in 47cd484b8657d54551faa2b29fcb65687295e4ed

emayevski:
> Wow, thank you very much for the contributions, it will have things helpful for many users. Can you fix the conflicts 

Interesting - the conflict was not indicated during the commit or a pull request. Nevermind, I've addressed it.

emayevski:
> @Bykiev can you help me review this? Lots of information to take on, but LGTM except for the complexity of `AdvancedExpressionOptions`, idk if this can be merged at `ExpressionOptions`.

ExpressionOptions is an enum, and fitting all of AdvancedExpressionsOptions is not possible. As for the Flags within AdvancedExpressionsOptions  - those flags that may or should work without AdvancedExpressionsOptions went into ExpressionOptions (I've added 4 or 5 flags there). Those that apply only to AdvancedExpressionsOptions had to go to AdvExpressionOptions and AdvancedExpressionsOptions.Flags anyway.

Bykiev (changes requested):

src/NCalc.Async/Visitors/AsyncEvaluationVisitor.cs
```diff
@@ -1,273 +1,398 @@
...
+
+            await OnUpdateParameterAsync(identifierName, parameterArgs);
+
+            if (!parameterArgs.UpdateParameterLists)
+            {
+                return value;
+            }
+
+            context.StaticParameters[context.Options.HasFlag(ExpressionOptions.LowerCaseIdentifierLookup) ? identifierName.ToLowerInvariant() : identifierName] = value;
+        }
```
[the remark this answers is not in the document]
emayevski: In the specific case of the option to use lowercase lookup, the function and variable names are expected to be ascii, so the possible cultural issues should not take place.  And it is simply more natural to have functions and variables written in lowercase than in uppercase - people normally type in lowercase. Also, let me point that the existing recommendations to achieve case-insensitive lookups are still valid.

Bykiev:
@gumbarros, I'd say we should rename `DontParseGuids` into `ParseGuids`, 'DontParseDates' into `ParseDates` and revert the logic.  I think we can release current branch as 5.5. Then we can merge this PR and release it as 6.0, because there are a lot of changes and new features. What do you think?

emayevski:
> @gumbarros, I'd say we should rename `DontParseGuids` into `ParseGuids`, 'DontParseDates' into `ParseDates` and invert the logic. 

In this case, it makes sense to add the default set of flags as a Default flag and use it where ExpressionOptions.None is passed now (in multiple places) so that compatibility is preserved. I didn't do this to keep the changes to the core code at a minimum.

gumbarros:
Hey @emayevski! Thanks a lot for the contribution, and sorry I haven’t been around, things have been really busy on my side with work and personal life. That said, this PR is too big and mixes multiple features, which isn’t a good practice, if I merge something of +8739 lines and -2739 at @jjconsulting I would be fired 😅. Can you plese split it into smaller PRs by feature? It makes reviewing manageable and is essential for a clear changelog. Thanks for understanding.

PR #447, opened 2025-07-13
refactor: Unify sync and async expression handling

### LINKED ISSUE #430: Simplify Architecture by Unifying Sync/Async Code Paths
To simplify the architecture, what about unifying the shared logic between `NCalcSync` and `NCalcAsync`?

We could merge the duplicated logic from both sync and async versions, and have the sync path simply execute the `ValueTask` synchronously, similar to this approach:
[https://github.com/sebastienros/fluid/blob/dc7c28a9d2e0beb47ed2f527d17c9286a2da08f3/Fluid/FluidTemplateExtensions.Sync.cs#L42](https://github.com/sebastienros/fluid/blob/dc7c28a9d2e0beb47ed2f527d17c9286a2da08f3/Fluid/FluidTemplateExtensions.Sync.cs#L42)

The sync version might be *slightly* slower, but the architectural benefits would be significant.

gumbarros:
This closes #430


gumbarros:
Features:
- Users can now use both sync and async from the same `Expression` or `ExpressionContext` instance
- Maintaince is much easier for us, we no longer need to duplicate code for both sync and async

BREAKING CHANGES (We should release this at a v6 and not v5.*):
- NCalcSync and NCalcAsync no longer exist. Consumers must install NCalc.Core.
- `EvaluateFunctionHandler` and `EvaluateParameterHandler` now return a `ValueTask`.
- `AsyncExpression` and all related async classes have been removed. Use `AsyncFunctions` and `AsyncDynamicParameters` properties within `Expression` to handle asynchronous operations, including CPU-bound and I/O-bound workloads.

gumbarros:
@Bykiev this is ~300ns slower and allocates a few more bytes than the other PR, I think it's negligible for any real-world use. Users will notice a large asset file in a web app or a UI freeze in a desktop app way before they notice this 😁

Bykiev:
> @Bykiev this is ~300ns slower and allocates a few more bytes than another PR, I think it's negligible for any real-world use. Users will notice a large asset file in a web app or a UI freeze in a desktop app way before they notice this 😁
> 
> Maybe with the codebase simpler, we can think in more ways to improve perf

I believe it's ok, because it's hard to measure such a small diff, even with BenchmarkDotNet the results can't be 100% accurate

emayevski:
@gumbarros  The approach you used to implementing Evaluate (via waiting for a result) can cause a deadlock in some environments. Yes, it is tempting to take a shortcut, but the right design is to implement two branches of code (it can be within one set of classes or in different classes as before).

Issue #367, opened 2024-11-27
Additional events (OnExpressionEvaluating, OnExpressionEvaluated, OnFunctionEvaluating, OnFunctionEvaluated)


amos-ts:
**Is your feature request related to a problem? Please describe.**
When expression string have a lot of operators or function and result in deep logical expression tree (logical expression) it could result in StackOverflowException during logical expression evaluation.

i.e. expression like if(condition, 1, if(confition2, 2, if(condition3, 3, if(.....)))) where we will repeat if in else block for i.e. 5K times will result StackOverflowExcpeption is expression input parameters results in latest if else block
This nested if could be solved with using if ifs() function, but same issue will happen if we have deep nested functions calls with any function.

The same StackOverflowException result if you have expression 1+2+3+4+....+5000+5001

There is separate issue where expression parser (during expression string parsing into LogicalExpression - where expression have huge number of operators or nested function calls,) could result with StackOverflowException. But I will report that as separate issue (or feature as I had some idea how caller could be notified of potential deep logical expression).

**Describe the solution you'd like**
Add events
1) OnExpressionEvaluating
This event will be called before any LogicalExpression evaluation (Binary, Unart, Function, etc.).
Arguments could be:
- LogicalExpression that is about to be evaluated
- Result - If set at OnExpressionEvaluating, then do not evaluate LogicalExpression and use that result

2) OnExpressionEvaluated
This event will be called after any LogicalExpression is evaluated (Binary, Unart, Function, etc.) and will contains evaluated expression result
Arguments could be:
- LogicalExpression that is evaluated
- Result - represent expression result. Could be used in handler to override result if needed

3) OnFunctionEvaluating
This event will be called before any Function is evaluated (built-in or custom functions).
Arguments could be:
- Same arguments and sender as it is presented in existing EvaluateFunction/EvaluateFunctionAsync events
- If arg.Result is set in this handler then do not evaluate function

4) OnFunctionEvaluated
This event will be called after any Function is evaluated (built-in or custom functions) and will contains evaluated function result.
Arguments could be:
- Same arguments and sender as it is presented in existing EvaluateFunction/EvaluateFunctionAsync events
- arg.Result could be used to override function result.

**Describe alternatives you've considered**

**Additional context**
With these new events we could track depth of expression tree during expression evaluation.
Also user of library could add expression caching. I.e. We could cache previously evaluated functions and if we get same function call with same arguments we could return cached value at OnFunctionEvaluating. Same apply for any other expression caching.

[...]

PR #450, opened 2025-07-17
Parlot - use Parser.Parse instead of Parser.TryParse

Bykiev:
Improve Parlot parser performance by using Parser.Parse, instead of Parser.TryParse.

This PR is based on Parlot's benchmark:

BenchmarkDotNet v0.14.0, Windows 11 (10.0.26100.2314)
12th Gen Intel Core i7-1260P, 1 CPU, 16 logical and 12 physical cores
.NET SDK 9.0.100
  [Host]   : .NET 9.0.0 (9.0.24.52809), X64 RyuJIT AVX2
  ShortRun : .NET 9.0.0 (9.0.24.52809), X64 RyuJIT AVX2

Job=ShortRun  IterationCount=3  LaunchCount=1
WarmupCount=3

| Method              | Mean        | Error       | StdDev    | Ratio | Allocated | Alloc Ratio |
|-------------------- |------------:|------------:|----------:|------:|----------:|------------:|
| ParlotRawSmall      |    201.4 ns |    53.24 ns |   2.92 ns |  0.49 |     304 B |        0.44 |
| ParlotFluentSmall   |    460.2 ns |   206.97 ns |  11.34 ns |  1.13 |     688 B |        1.00 |

2x time difference, it's seems the overhead is caused by `out` params and try/catch block and JITting is not applied.
But our benchmarks are confusing:

BenchmarkDotNet v0.15.1, Linux Ubuntu 24.04.2 LTS (Noble Numbat)
.NET SDK 9.0.302
  [Host]     : .NET 8.0.18 (8.0.1825.31117), X64 RyuJIT AVX2
  DefaultJob : .NET 8.0.18 (8.0.1825.31117), X64 RyuJIT AVX2

Before:
| Method                   | Mean      | Error     | StdDev    | Rank | Gen0   | Gen1   | Allocated |
|------------------------- |----------:|----------:|----------:|-----:|-------:|-------:|----------:|
| SimpleParlotExpression   |  6.596 us | 0.0244 us | 0.0217 us |    1 | 0.0610 |      - |   1.05 KB |
| AdvancedParlotExpression | 16.728 us | 0.0797 us | 0.0707 us |    3 | 0.1526 |      - |   2.98 KB |
| LambdaWithoutCompilation |   0.3696 ns | 0.0062 ns | 0.0055 ns |    1 |      - |         - |
| LambdaWithCompilation    | 702.3531 ns | 6.1587 ns | 5.7608 ns |    2 | 0.0248 |     424 B |
| Evaluate                 | 743.7670 ns | 7.3436 ns | 6.8692 ns |    3 | 0.1059 |    1776 B |

After:
| Method                   | Mean      | Error     | StdDev    | Rank | Gen0   | Gen1   | Allocated |
|------------------------- |----------:|----------:|----------:|-----:|-------:|-------:|----------:|
| SimpleParlotExpression   |  6.417 us | 0.0178 us | 0.0158 us |    1 | 0.0610 |      - |   1.05 KB |
| AdvancedParlotExpression | 16.536 us | 0.0220 us | 0.0172 us |    3 | 0.1526 |      - |   2.98 KB |
| LambdaWithoutCompilation |   0.3690 ns |  0.0057 ns | 0.0048 ns |    1 |      - |         - |
| LambdaWithCompilation    | 722.0903 ns |  3.9702 ns | 3.5195 ns |    2 | 0.0248 |     424 B |
| Evaluate                 | 767.6887 ns | 10.9282 ns | 9.1255 ns |    3 | 0.1059 |    1776 B |

So, some benchmarks shows improved result with Parser.Parse method, but some of them degraded. Seems this PR doesn't make any sense...

### DIFF
--- a/src/NCalc.Core/Parser/LogicalExpressionParser.cs
+++ b/src/NCalc.Core/Parser/LogicalExpressionParser.cs
@@ -469,15 +469,21 @@ public static class LogicalExpressionParser
     {
         var parser = GetOrCreateExpressionParser(context.CultureInfo);
 
-        if (parser.TryParse(context, out var result, out var error))
+        try
+        {
+            var result = parser.Parse(context);
+            if (result == null)
+            {
+                string message = $"Error parsing the expression at position {context.Scanner.Cursor.Position}";
+                throw new NCalcParserException(message);
+            }
+
             return result;
-
-        string message;
-        if (error != null)
-            message = $"{error.Message} at position {error.Position}";
-        else
-            message = $"Error parsing the expression at position {context.Scanner.Cursor.Position}";
-
-        throw new NCalcParserException(message);
+        }
+        catch (ParseException ex)
+        {
+            string message = $"{ex.Message} at position {ex.Position}";
+            throw new NCalcParserException(message);
+        }
     }
 }


emayevski:
@Bykiev To me, the change looks appealing for non-performance reasons: it gives a bit more control over the possible error (unfortunately, the errors produced now are very confusing for a human in some cases). As for the performance - it will only be noticeable in bulk processing. But even then, the difference is minute and may be explained by some external factors (you were not doing measurements concurrently in parallel on two cores, were you?).

Bykiev:
> @Bykiev To me, the change looks appealing for non-performance reasons: it gives a bit more control over the possible error (unfortunately, the errors produced now are very confusing for a human in some cases). 

Unfortunately, smart error handling will lead to poor performance, we can introduce some additional events for debugging purpose, as asked here: https://github.com/ncalc/ncalc/issues/367

emayevski:
> @Bykiev  Unfortunately, smart error handling will lead to poor performance, we can introduce some additional events for debugging purpose, as asked here: #367

I wrote only about the part you changed, not about every error being reported on the fly. There is no difference in performance there unless you parse tons of expressions that contain errors. On-the-fly reporting of the operations is a different story.

PR #353, opened 2024-10-30
Improve SerializationVisitor performance

Bykiev:
This PR improves `SerializationVisitor` performance. 

I did some benchmarking and `StringBuilder` usage in some case is not the best solution, here some benchmark results:

```
[MemoryDiagnoser]
public class TestComparision
{
	public const string identifier = "identifier";

	[Benchmark(Baseline = true)]
	public string TestStringBuilder()
	{
		var result = new StringBuilder();
		result.Append('(');
		result.Append(identifier);

		// trim spaces before adding a closing paren
		while (result[^1] == ' ')
			result.Remove(result.Length - 1, 1);

		result.Append(") ");
		return result.ToString();
	}

	[Benchmark]
	public string TestStringConcat()
	{
		return "(" + identifier.TrimEnd(' ') + ") ";
	}

	[Benchmark]
	public string TestStringInterpolation()
	{
		return $"({identifier.TrimEnd(' ')}) ";
	}
}
```

BenchmarkDotNet v0.14.0, Windows 10 (10.0.19045.5011/22H2/2022Update)
Intel Core i3-6100U CPU 2.30GHz (Skylake), 1 CPU, 4 logical and 2 physical cores
.NET SDK 8.0.206
  [Host]     : .NET 6.0.35 (6.0.3524.45918), X64 RyuJIT AVX2
  DefaultJob : .NET 6.0.35 (6.0.3524.45918), X64 RyuJIT AVX2

| Method                  | Mean     | Error    | StdDev   | Median   | Ratio | RatioSD | Gen0   | Allocated | Alloc Ratio |
|------------------------ |---------:|---------:|---------:|---------:|------:|--------:|-------:|----------:|------------:|
| TestStringBuilder       | 50.02 ns | 1.114 ns | 1.598 ns | 49.23 ns |  1.00 |    0.04 | 0.0969 |     152 B |        1.00 |
| TestStringConcat        | 26.80 ns | 0.228 ns | 0.190 ns | 26.74 ns |  0.54 |    0.02 | 0.0306 |      48 B |        0.32 |
| TestStringInterpolation | 27.64 ns | 0.625 ns | 1.647 ns | 26.73 ns |  0.55 |    0.04 | 0.0306 |      48 B |        0.32 |

  
```
[MemoryDiagnoser]
public class TestComparision
{
	public const string identifier = "identifier";

	[Benchmark(Baseline = true)]
	public string TestStringBuilderLoop()
	{
		var result = new StringBuilder();

[...]

PR #451, opened 2025-07-21
Improve SerializationVisitor perf

Bykiev:
As the ObjectPool was reverted in #409 due to dependency issue the PR from https://github.com/ncalc/ncalc/pull/353/commits/eda56a3b549a16ea320fa676dedb102653774ece is applied again.

### DIFF
--- a/src/NCalc.Core/Visitors/SerializationVisitor.cs
+++ b/src/NCalc.Core/Visitors/SerializationVisitor.cs
@@ -131,7 +131,7 @@ public class SerializationVisitor : ILogicalExpressionVisitor<string>
         if (expression is ValueExpression valueExpression)
             return valueExpression.Accept(this);
 
-        string result = expression.Accept(this);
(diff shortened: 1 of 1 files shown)


Bykiev:
@emayevski, can you please review the PR? Unfortunately can't add you to reviewers (seems only contributors can be added to review)

emayevski:
@Bykiev  Getting rid of StringBuilder probably makes sense, just let me ask one thing - in the Visit(TernaryExpression) method with five concatenated strings, wouldn't it be faster to call string.Concat instead of just "+"? I don't know if the compiler internally does this (I'd need to spend an hour to figuring this out) but if you made measurements, maybe you have an easy way to compare these approaches.

emayevski:
@Bykiev, that's interesting, thank you. I suspected that Concat is used under the hood for performance benefits but didn't have a chance to check.

PR #453, opened 2025-07-22
Move IsReal check in MathHelper

Bykiev:
Move IsReal check in MathHelper


Bykiev:
@emayevski, can you please review the PR? Unfortunately can't add you to reviewers (seems only contributors can be added to review)

emayevski:
@Bykiev ...and now I see that I can shift the check for null above other checks as well.

gumbarros:
Can we merge it? Or too dangerous?

Bykiev:
I believe it's safe to merge this PR, but theoretically there could be other issues when using null parameters

gumbarros:
> I believe it's safe to merge this PR, but theoretically there could be other issues when using null parameters

As a wise man once told me: **you only live once**. Let’s merge and test it in our projects 😅

PR #454, opened 2025-07-24
Refactor Like/NotLike functions

Bykiev:


Bykiev:
I see no reason for it, every built-in function can be overridden by user

gumbarros:
LGTM

PR #456, opened 2025-08-03
Move lambda compilation to `NCalc.LambdaCompilation`

### LINKED ISSUE #455: Move Lambda compilation to another assembly
For people like me who don't use lambda compilation, `FastExpressionCompiler` becomes an unnecessary dependency:

- Adds more .dlls to the deploy
- Creates extra noise in the IDE (see screenshot)
- No benefit for projects that don’t use the feature

Can we either make this optional or allow an easy way to disable/remove it?
Image

gumbarros:
This closes #455


src/Plugins/NCalc.MemoryCache/README.md
```diff
@@ -1,4 +1,4 @@
-# NCalc.MemoryCache
+# NCalc.LambdaCompilation
```
Bykiev, line 1: Can you please fix both readme files?

src/Plugins/NCalc.LambdaCompilation/README.md
```diff
@@ -0,0 +1,6 @@
+# NCalc.MemoryCache
```
Bykiev, line 1: Looks like should be NCalc.LambdaCompilation

Bykiev (changes requested):

Bykiev (approved):
Thank you, good job!

PR #461, opened 2025-08-22
Use static lambdas in parser where possible

Bykiev:
Use static lambdas in parser where possible


Bykiev:
@emayevski, can you please review this PR?

emayevski:
> @emayevski, can you please review this PR?

Looks fine to me. There are cases within a parser though, where lambdas need external variables - I remember that I needed to remove "static" from some of the lambdas in the parser. Time for me to revisit this code too.

PR #463, opened 2025-08-25
Lambda compilation missing support for LIKE/NOT LIKE operators

zkenda:
Added support for LIKE/NOT LIKE operators when using lambda compilation


gumbarros:
@zkenda thanks for the contribution.
image

Can you fix the conflicts?

PR #464, opened 2025-08-25
Lambda compilation missing support for IN/NOT IN operators

zkenda:
Added missing support for IN/NOT IN operators when compiling to lambda


gumbarros:
Thanks for the contribution!

PR #466, opened 2025-08-29
Add dynamic argument separator support for function arguments

attiqerehman:
Introduce support for dynamic argument separators in function expressions, allowing customization of the separator character. Include tests to verify functionality with both comma and semicolon separators, as well as handling of incorrect separators.


emayevski:
Choosing a comma as a default separator has the following danger: in cultures where comma is a decimal separator (large parts of Europe, for example), a list (1,2,3) will be parsed as two elements "1,2" and "3".

gumbarros:
@attiqerehman I saw your PR just now, nice try, it probably fails, because we need multiple parse instances support, after the parser is created (static ctor) it's impossible to change. We would need a class like `LogicalExpressionParserOptions` and pass at the ctor, change the parameter and cache the parser.

attiqerehman:
I will have an another go at it.

PR #477, opened 2025-09-24
Mark functions in SerializationVistor as virtual

randellhodges:
This marks all the methods as virtual to allow the methods to be overridden in subclasses like it was back in the 3.x versions.

In my project, I subclass the SerializationVisitor to cut down on brackets, and provide a means to parse and rename functions and identifiers. By marking the methods as virtual, I can continue to subclass without having to copy the code for everything.


Bykiev (approved):
LGTM, thank you!

Issue #479, opened 2025-10-14
Operator '+' can't be applied to operands of types 'int' and 'ulong'


NTDLS:
```csharp
//Works as expected:
var passExp = new NCalc.Expression("v1+(v2^v3)");
passExp.Parameters["v1"] = (double)1;
passExp.Parameters["v2"] = (double)2;
passExp.Parameters["v3"] = (double)3;
var passRes = passExp.Evaluate();

//Fails with: System.InvalidOperationException: 'Operator '+' can't be applied to operands of types 'int' and 'ulong''
var failExp = new NCalc.Expression("1+(v3^v4)");
failExp.Parameters["v3"] = (double)2;
failExp.Parameters["v4"] = (double)3;
var failRes = failExp.Evaluate();

//Fails with: System.InvalidOperationException: 'Operator '+' can't be applied to operands of types 'int' and 'ulong''
var failExp2 = new NCalc.Expression("v1+(v2^v3)");
failExp2.Parameters["v1"] = 1;
failExp2.Parameters["v2"] = 2;
failExp2.Parameters["v3"] = 3;
var failRes2 = failExp2.Evaluate();
```

👀  If it helps, these errors do not occur with https://www.nuget.org/packages/CoreCLR-NCalc

<img width="915" height="243" alt="Image" src="https://github.com/user-attachments/assets/18d9c6aa-cdf6-496a-82fd-ae5bc2dd9d67" />

<img width="496" height="225" alt="Image" src="https://github.com/user-attachments/assets/679d8b66-8914-4f3d-b44e-6c2c6a63e99e" />

<img width="1002" height="125" alt="Image" src="https://github.com/user-attachments/assets/e1494497-b7a1-4bd0-95e0-b22a391c57ec" />

Bykiev:
Makes sense, I'll research it at the evening, thanks!

Bykiev:
This should be fixed in v5.8.0

PR #481, opened 2025-10-14
Allow math operations with ulong types

Bykiev:
This PR adds support for using math operations with ulong types if possible.

credits to @emayevski

References #479

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -10,7 +10,7 @@
   </PropertyGroup>
   <ItemGroup>
     <PackageReference Include="Nerdbank.GitVersioning">
-      <Version>3.7.115</Version>
+      <Version>3.8.118</Version>
       <PrivateAssets>all</PrivateAssets>
     </PackageReference>
     <PackageReference Include="PolySharp" Version="1.15.0">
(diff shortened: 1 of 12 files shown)


emayevski (commented):
I would suggest verifying border cases such as _(ushort) 65535 vs (short) -32767_ (both must be converted to int first, then an operation must be performed. This is something I meant by "needs more checking" in the comment - I am not completely confident in that ConvertToHighestPrecision function. Yes it should work but does it handle such border cases properly?

emayevski:
> @emayevski, nice catch! Seems the manual conversion between the types is the only correct solution

Not exactly. I have refactored the code of my version and implemented proper type conversion within ConvertToHighestPrecision, getting rid of ExecuteXYZFunction methods. Please see https://github.com/Allied-Bits-Ltd/ncalc/blob/1d43b4e8adbdaa6c5435411ac61784282e1f7909/src/NCalc/Helpers/MathHelper.cs . There's quite a number of changes in that commit ( I have refactored checks for null and other minor stuff), but you are interested mostly in the code of ConvertToHighestPrecision.

emayevski:
@Bykiev  This may be the second issue I mentioned initially — bit-extending operations on signed values may work wrong. I will check this case now.

Bykiev:
@emayevski, I've reverted the changes with `ConvertToHighestPrecision` and I'm going to release a bugfix. We can work on your change in another PR, but I think there are some more tests needed.

emayevski:
@Bykiev  Upon further thinking, I see even more serious problem. 

int a = Int32.MinValue;
int b = Int32.MinValue;

"a - b" will cause numeric underflow. Here, we clearly need bit expansion, but how would one determine that such expansion is even necessary? It may be more reasonable to always expand or expand if the operation failed with an overflow exception. 

And this problem doesn't depend on my changes, BTW.

Bykiev:
> It may be more reasonable to always expand

I think it may be a good solution, but needs testing

emayevski:
FWIW I made more improvements related to overflows and added a bunch of tests of border cases. They all pass, so I made another update. Here's the commit you may be interested in: https://github.com/Allied-Bits-Ltd/ncalc/commit/2c0953248281a24337910cab386d17d197f41072 .  The changes are relatively complicated - we first try to perform the operation as is, and if it overflows, we request the bit expansion even if the type is the same and repeat the operation until it is successful or there is nowhere more to expand. It is important that OverflowProtection is enabled in the options: by default, it is disabled and the code silently produces a wrong result. I would say that this approach sucks, but changing it is a major breaking change that can cause problems for users.

emayevski:
@Bykiev  The idea of expanding the type if one of the parameters is signed and another one is unsigned makes some sense, but you have left aside some overflow cases, which my changes address. Just run the tests I added with your code, and you'll see what works and what does not. I don't see any other problems in your code, but it's Sunday evening, and automated tests can uncover more than manual inspection.

gumbarros (approved):

gumbarros:
> @gumbarros, can you review this PR please?

Sorry for the late approval, good work.

PR #484, opened 2025-10-21
Add LongAsDefault option

### LINKED ISSUE #411: Expression 10000000*1000 gives incorrect result. This looks to be due to internal overflow.
Evaluating the simple numeric expressions below results in an internal overflow for lines 2,3 and 4 and results in wrong result. Adding the "OverflowProtection" option causes an exception for the expression items 2,3, and 4 below.

    Expression = result 
1. 1000000*1000 =1000000000
2. 10000000*1000=1410065408
3. 100000000*1000=1215752192
4. 1000000000*1000=-727379968
5. 10000000000*1000=10000000000000

This is known to be happening in version 5.4.2 and 5.4.1, other versions are untested.

Bykiev:
Add support for `LongAsDefault` option

Closes #411

### DIFF
--- a/src/NCalc.Core/NCalc.Core.csproj
+++ b/src/NCalc.Core/NCalc.Core.csproj
@@ -35,6 +35,6 @@
         <PackageReference Include="ExtendedNumerics.BigDecimal" Version="3001.1.0.233" />
         <PackageReference Include="Microsoft.Extensions.Logging.Console" Version="9.0.10" />
         <PackageReference Include="Microsoft.Extensions.Logging.TraceSource" Version="9.0.10" />
-        <PackageReference Include="Parlot" Version="1.5.0" />
+        <PackageReference Include="Parlot" Version="1.5.1" />
     </ItemGroup>
 </Project>


src/NCalc.Core/Parser/LogicalExpressionParser.cs
```diff
@@ -114,7 +114,14 @@ private static Parser<LogicalExpression> CreateExpressionParser(CultureInfo cult
...
             .AndSkip(Not(OneOf(Terms.Text("."), Terms.Text("E", true))))
-            .Then<LogicalExpression>(static d => new ValueExpression(d));
+            .Then<LogicalExpression>(static (ctx, val) =>
+            {
+                bool useLong = ((LogicalExpressionParserContext)ctx).Options.HasFlag(ExpressionOptions.LongAsDefault);
+                if (useLong)
+                    return new ValueExpression((long)val);
+
+                return new ValueExpression(val);
+            });
```
gumbarros: I think it's possible to optimize performance here avoiding a cast by using `Terms.Number<long>` and doing the if at parser level, but maybe I'm wrong.

Bykiev:
Converted to draft, going to use a `Switch` parser as more suitable

src/NCalc.Core/Helpers/MathHelperOptions.cs
```diff
@@ -30,6 +30,12 @@ public bool AllowCharValues
         get => options.HasFlag(ExpressionOptions.AllowCharValues);
     }
 
+    public bool LongAsDefault
+    {
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        get => options.HasFlag(ExpressionOptions.LongAsDefault);
```
lahma: if I remember correctly the `HasFlag` can be notoriously inefficient depending on the runtime... if you target netstandard you might want to consider just bit operations

Bykiev: Hi, nice to see your here!

This doesn't seems be an issue, because we're running benchmark on .net 8.0, and the first PR with adding this new option doesn't have any perf impact: https://github.com/ncalc/ncalc/actions/runs/18692621173/job/53301871481

Seems to be a`Select` parser issue

gumbarros (approved):

src/NCalc.Core/Parser/LogicalExpressionParser.cs
```diff
@@ -364,12 +364,19 @@ private static Parser<LogicalExpression> CreateExpressionParser(CultureInfo cult
...
+            return OneOf(intNumber, longNumber);
+        });
+
+        // primary => GUID | NUMBER | identifier | DateTime | string | function | boolean | groupExpression | list ;
         var primary = OneOf(
             guid,
             hexOctBinNumber,
-            intNumber,
-            longNumber,
+            integralNumber,
```
sebastienros: `Select` is not `ISeekable` that could prevent the whole `primary` parser from using a lookup and kill perf (or at least the nice free optimization parlot will find out).

You can try to use `Lookup` after the Select to give the list of valid chars since the expected chars is known.

`Select(...).Lookup(integerNumber, longNumber)`

sebastienros: The other option (which I think I already suggested) is to only parse `Number<long>` but use `Then()` to convert to `int` based on the option of now. One line, much simpler, no Select

src/NCalc.Core/Parser/LogicalExpressionParser.cs
```diff
@@ -364,12 +364,19 @@ private static Parser<LogicalExpression> CreateExpressionParser(CultureInfo cult
 
         var guid = OneOf(guidWithHyphens, guidWithoutHyphens);
 
-        // primary => GUID | NUMBER | identifier| DateTime | string | function | boolean | groupExpression | list ;
+        var integralNumber = Select<LogicalExpressionParserContext, LogicalExpression>((ctx) =>
+        {
+            if (ctx.Options.HasFlag(ExpressionOptions.LongAsDefault))
+                return longNumber;
+
+            return OneOf(intNumber, longNumber);
```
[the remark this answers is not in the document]
Bykiev: > I think evaluation the flag for each number won't matter much in the end, but yes this could be measured with and without the chech.
> 
> and somewhere our of this lambda:
> 
> `var intOrLong = OneOf(intNumber, longNumber)`
> 
> like it's done for `decimalOrDoubleNumber` on line 151.

That did a trick, thank you!

PR #486, opened 2025-10-27
Improve double parser

Bykiev:
This PR removes casting decimal value to double and uses a double parser instead.


emayevski (approved):
Makes sense - the previous version was not aligned with the name of the option.

PR #508, opened 2025-11-23
Add CancellationToken support

### LINKED ISSUE #502: Add cancellation token support
I want to add cancellation token support in v 5.9.0, but waiting for a new Parlot release (https://github.com/sebastienros/parlot/pull/271).

@sebastienros, do you have any ETA on this?

Bykiev:
This PR adds CancellationToken support. This is a breaking change

Closes #502

### DIFF
--- a/docs/articles/cancellation.md
+++ b/docs/articles/cancellation.md
@@ -14,7 +14,7 @@ var cts = new CancellationTokenSource();
 var expression = new AsyncExpression("getUserName(1) == 'admin'");
 expression.Functions["getUserName"] = async args =>
 {
-    var id = (int)(await args[0].EvaluateAsync(cts.Token)!)!;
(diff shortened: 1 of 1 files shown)


Bykiev:
I believe it's a non-standard way in .NET, it's a subscriber responsibility to handle cancellation. We can add it in the args (AsyncFunctionArgs for ex.) if you wish. What do you think?

gumbarros:
> I believe it's a non-standard way in .NET, it's a subscriber responsibility to handle cancellation. We can add it in the args (AsyncFunctionArgs for ex.) if you wish. What do you think?

Perfect, that's what I mean. I agree with you, the cancellation token should be at the function parameters and the subscriber pass the token forward.

Bykiev:
> We should release the next version as v6 because of the breaking change?

There is no much breaking, I believe we can release it as v5.9

PR #510, opened 2025-11-23
Optimize In.Contains method

Bykiev:
Avoid `string.IsNullOrEmpty` double check

### DIFF
--- a/src/NCalc.Core/Helpers/EvaluationHelper.cs
+++ b/src/NCalc.Core/Helpers/EvaluationHelper.cs
@@ -72,11 +72,8 @@ public static class EvaluationHelper
 
         var leftValueString = Convert.ToString(leftValue, CultureInfo.InvariantCulture);
 
-        if (string.IsNullOrEmpty(leftValueString) && string.IsNullOrEmpty(rightValue))
-            return true;
-
         if (string.IsNullOrEmpty(leftValueString))
-            return false;
+            return string.IsNullOrEmpty(rightValue);
 
         return rightValue.Contains(leftValueString);
     }


gumbarros:
> I've used sharplab.io to compare ASM:
> 
> [original method](https://sharplab.io/#v2:C4LghgzgtgPgAgJgIwFgBQcAMACOSB0A4gDYD2ARmMQJYBeYw1pAdgNzrpwDMuC2AwtgDe6bGOwAHAE7UAbgwCmuJADZs5UqWICWwMNWYQAFBQBWCgMbAA/NmIKAZsABqVAK4KANMpwyA5gAWLu4KAJSi4iJo4jHY8lJ2jsHEHgDKwDLMftgAvDrMsgpSwPgAKqTpmX5G9k6uKV4CbsTAblIKAJLMDqT4XfHUYMzA/M2t7aHsaBGx2NQO2EZ4mH0QAHLNxADyUgCiUBLAAJ41SfVpGQZ+odgAZLc+qxvE23sHx0b+Qedh4dGzsTgAHZsBkPFMZrF5otlk9Njt9ocTrVkhcqqE/gDASCHFQIAoIf9ZsDsF9UQp8PxdPpDKc6iFKldJpCAL7oFlAA=) [modified](https://sharplab.io/#v2:C4LghgzgtgPgAgJgIwFgBQcAMACOSB0A4gDYD2ARmMQJYBeYw1pAdgNzrpwDMuC2AwtgDe6bGOwAHAE7UAbgwCmuJADZs5UqWICWwMNWYQAFBQBWCgMbAA/NmIKAZsABqVAK4KANMpwyA5gAWLu4KAJSi4iJo4jHY8lJ2jsHEHgDKwDLMftgAvDrMsgpSwPgAKqTpmX5G9k6uKV4CbsTAblIKAJLMDqT4XfHUYMzA/M2t7aHsaBGx2NQO2EZ4mH0QAHLNxADyUgCiUBLAAJ41SfVpGQZ+oeHRs7FwAOw+qxvE23sHx0b+QedhUxmD2ev2SHnw/F0+kMpzqIUqV0mQIAvuhkUA===)
> 
> I have no much experince with it, that's why it's from Qwen AI:
> 
> **Register Usage Optimization:**
> 
> In the first version, 3 registers (esi, ebx, eax) are used for temporary values In the second version, the number of registers is reduced to 2 (ebx, eax), which is more efficient
> 
> **Differences in Null Value Handling Logic:**
> 
> In the first version: after checking test eax, eax and jumping to je short L001d, there's an additional string length check via cmp dword ptr [eax+4], 0 In the second version: the logic is simplified - when eax is null, it immediately jumps to process ebx
> 
> **Processing Efficiency:**
> 
> The second version uses sete al and movzx eax, al to set the boolean result, which is more compact The first version uses separate branches for each case, increasing the number of instructions
> 
> **Code Size:**
> 
> First version: 5Dh (93) bytes Second version: 49h (73) bytes Conclusion: The second version represents a more optimized code with fewer instructions, more efficient register usage, and simplified logic for handling edge cases. This indicates more sophisticated compiler optimization in the second case.
> 
> Of cource, the perf difference is negligible, but it's worth than nothing :)

Wow, I would never check this impact, nice details. At a web app at the cloud this will probably be negligent, but for people running NCalc inside DOOM or some very low level app, this maybe be useful 😄.

PR #512, opened 2025-11-23
Refactor `EvaluationHelper` to support generic type context

gumbarros:
Refactored the `EvaluationHelper` class to use a generic type parameter for context, improving performance, because there is no more type casting of `ExpressionContext` to `ExpressionContextBase` for example.


Bykiev (approved):
Awesome!

PR #516, opened 2025-11-23
Fix docs build

Bykiev:


Bykiev:
Seems DocFX doesn't support C#14 yet
https://github.com/dotnet/docfx/issues/10808

PR #519, opened 2025-11-28
Add factorial unary expression support

### LINKED ISSUE #490: Support for factorial n!
**Is your feature request related to a problem? Please describe.**
No.

**Describe the solution you'd like**
```csharp
var expression = new Expression("5!");
Debug.Assert(120 == expression.Evaluate());
```

**Describe alternatives you've considered**
I manually support the factorial operator in my program by doing some regex replace on the expression string before creating an `Expression` object with it.

**Additional context**
I'm not sure where factorial falls in the order of operations, so I'm not certain how this expression should be evaluated, for example:
```
"5!=0"
```
`true` because 5 does equal 0? Or `false` because 120 does not equal 0?

gumbarros:
This closes #490

### DIFF
--- a/src/NCalc.Core/Parser/LogicalExpressionParser.cs
+++ b/src/NCalc.Core/Parser/LogicalExpressionParser.cs
@@ -152,7 +152,6 @@ public static class LogicalExpressionParser
         });
 
         var argumentSeparatorTerm = Terms.Char(argumentSeparator);
-        var comma = Terms.Char(',');
         var divided = Terms.Text("/");
         var times = Terms.Text("*");
         var modulo = Terms.Text("%");


docs/articles/operators.md
gumbarros: @Bykiev I don't like how the docs is looking, maybe it's better to abort factorial at parser and change to a built-in function?

gumbarros: Or re-write the doc

src/NCalc.Core/Helpers/MathHelper.cs
```diff
@@ -485,4 +485,51 @@ private static void CheckOverflow(dynamic value)
...
+        var r = one;
+
+        for (var i = one + one; i <= n; i++)
+            r *= i;
+
+        return r;
+    }
+#else
+    private static dynamic CalculateFactorial(dynamic n)
+    {
```
Bykiev: Can you please add negative value check to be more consistent with INumber variant?

gumbarros: Sure

src/NCalc.Core/Helpers/MathHelper.cs
```diff
@@ -485,4 +485,51 @@ private static void CheckOverflow(dynamic value)
                 throw new OverflowException("Arithmetic operation resulted in an overflow");
         }
     }
+
+    public static object Factorial(object? result)
+    {
+        switch (result)
```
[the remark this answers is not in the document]
Bykiev: I can't reproduce it on my machine, all works as expected:

```
public static object Factorial(object? result)
{
    return result switch
    {
        int v => CalculateFactorial(v),
        long v => CalculateFactorial(v),
        float v => CalculateFactorial(v),
        double v => CalculateFactorial(v),
        decimal v => CalculateFactorial(v),
        BigInteger v => CalculateFactorial(v),
        _ => throw new ArgumentException("Unsupported numeric type.", nameof(result)),
    };
}
```

Maybe try on CI?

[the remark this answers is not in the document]
gumbarros: As CI is working I see no problem. I also checked and my machine is at a pre-release version of .NET 10, so makes sense.

Issue #522, opened 2025-12-02
NCalc 5.9.0 crash when evaluating a simple expression when executed in AOT


vpenades:
**Describe the bug**

A simple expression crashes when executed in AOT, it works in normal debug and release.

**Example Expression**

```c#
namespace NcalcExpressionCrash
{
    internal class Program
    {
        static void Main(string[] args)
        {
            var expr = new NCalc.Expression("-100");         

            var val = expr.Evaluate();            

            Console.WriteLine(val.ToString());
        }        
    }
}
```

- Create a console program for Net.10, using NCalcSync 5.9.0, 
- Enable PublishAOT
- Publish
- run the published binary

The expected output is -100

The exception I get is:

```
Unhandled exception. System.ArgumentNullException: Value cannot be null. (Parameter 'key')
   at System.Collections.Generic.Dictionary`2.TryInsert(TKey, TValue, InsertionBehavior) + 0x275
   at Microsoft.CSharp.RuntimeBinder.ExpressionTreeCallRewriter.VisitSAVE(ExprBinOp) + 0x4a
   at Microsoft.CSharp.RuntimeBinder.Semantics.ExprVisitorBase.Dispatch(Expr) + 0x352
   at Microsoft.CSharp.RuntimeBinder.Semantics.ExprVisitorBase.VisitChildren(Expr) + 0x66
   at Microsoft.CSharp.RuntimeBinder.Semantics.ExprVisitorBase.Dispatch(Expr) + 0x38e
   at Microsoft.CSharp.RuntimeBinder.ExpressionTreeCallRewriter.Rewrite(ExprBinOp, Expression[]) + 0x62
   at Microsoft.CSharp.RuntimeBinder.RuntimeBinder.BindCore(ICSharpBinder, Expression[], DynamicMetaObject[], DynamicMetaObject&) + 0x15a
   at Microsoft.CSharp.RuntimeBinder.RuntimeBinder.Bind(ICSharpBinder, Expression[], DynamicMetaObject[], DynamicMetaObject&) + 0x4f
   at Microsoft.CSharp.RuntimeBinder.BinderHelper.Bind(ICSharpBinder, RuntimeBinder, DynamicMetaObject[], IEnumerable`1, DynamicMetaObject) + 0x4b6
   at Microsoft.CSharp.RuntimeBinder.CSharpBinaryOperationBinder.FallbackBinaryOperation(DynamicMetaObject, DynamicMetaObject, DynamicMetaObject) + 0x89
[...]

PR #523, opened 2025-12-03
Added basic AOT functionality

vpenades:
This PR is intended to provide basic AOT funcionality to NCalc.

It is neither as flexible nor as performant as the dynamic JIT code path, but it works, which is something that can save the day for those that are required to compile to AOT.

Right now the unit tests will run on the dynamic code path. To test the AOT code path,  set MathHelperOptions.IsDynamicCodeSupported to false in the source code and re-run the tests.

This PR tries to address #522


src/NCalc.Core/Helpers/BinaryOperators.AOT.cs
```diff
@@ -0,0 +1,202 @@
+﻿namespace NCalc.Helpers;
+
+/// <summary>
+/// Binary operators designed specifically for AOT execution
+/// </summary>
+static class BinaryOperatorsAOT
+{
+    // unchecked
+    public static object AddFunc(object a, object b)
```
gumbarros: Is possible here to use generics with <A,B>?

gumbarros (commented):
Thanks for your contribution

src/NCalc.Core/Helpers/BinaryOperators.AOT.cs
vpenades: No, I reviewed the generic math interfaces and they work in a way they require the types to be already casted to a known type beforehand. In other words, you can't avoid casting `object` to a specific type somewhere.

The only performance improvement I can think of is to have many versions of AddFunc, each one specifically designed for a certain type, as in:

```c#
public static object AddFunc_int(object a, object b) { return (int)a + (int)b; }
public static object AddFunc_float(object a, object b) { return (float)a + (float)b; }
```

And then, in MathHelper, use a.GetType() and have a switch-case block to return an operator function that assumes objects a and b are expected to be of a certain type.  This would effectivelly move the switch-case block out of the function to the expression solver. But that would also assume that the function, at that particular point, is always going to be called with object a and b being of that certain type, I really don't know enough of Ncalc to make that assumption.

Also I consider my PR as a starting point, there's room for improvement in future PRs

gumbarros:
> I'd say it's better than nothing. We need to run tests with AOT too, but xUnit is not AOT-compatible, we can migrate to TUnit: https://tunit.dev/docs/migration/xunit

I agree with the change, looks more modern than xunit.

sferhah:
This PR made it possible to ship my iOS app (Maui) in AOT mode. Looking forward to seeing it merged so I can drop the copied code :p

src/NCalc.Async/NCalc.Async.csproj
```diff
@@ -13,6 +13,10 @@
         <Description>NCalcAsync is a fully async port of the NCalc.NCalc is a fast and lightweight expression evaluator library for .NET, designed for flexibility and high performance. It supports a wide range of mathematical and logical operations. NCalc can parse any expression and evaluate the result, including static or dynamic parameters and custom functions.</Description>
     </PropertyGroup>
 
+  <PropertyGroup Condition="'$(TargetFramework)' == 'net8.0' OR '$(TargetFramework)' == 'net10.0'">
```
Bykiev: I don't like this condition, because we need to add new framework after each release. 

It should be better to use this one:

```
<PropertyGroup Condition="$([MSBuild]::IsTargetFrameworkCompatible('$(TargetFramework)', 'net8.0'))">
  <PublishTrimmed>true</PublishTrimmed>
</PropertyGroup>
```
Btw, should we use `IsAotCompatible` instead of `PublishTrimmed`?

emayevski:
> Sorry, but I fail to see where are the relevant code changes that makes your "fork" AOT ready, what have you changed from the original code that makes your code AOT ready?

You don't see it because the code is not a fork. The code is different there. You can check the history of commits to find the changes.

cuican6:
> or your contribution

Your project is of great help to me. I would like to know how the AOT progress is going. When is it expected to be officially available? Thank you

gumbarros:
We are now using TUnit for testing, what is now missing to this be merged?

Bykiev:
> We are now using TUnit for testing, what is now missing to this be merged?

First of all, we should use this configuration:

```
<PropertyGroup>
  <TargetFrameworks>net8.0;net9.0</TargetFrameworks>
  <IsAotCompatible>true</IsAotCompatible>
</PropertyGroup>
```

Then we need to add AOT tests on CI. After these changes we can merge this PR

vpenades:
@Bykiev I've added IsAotCompatible as a general property, but disabled it explicitly for NCalc.Lambdacompilation

Also removed Net9 in favour of net10, since net9 is at its end of support.

What's left is to be able to run the CI tests with and without AOT to check both paths, but I don't know how to do on the github workflow

gumbarros:
@vpenades try changing the CI at your fork like this:

```yaml
name: Build & Test

on:
  push:
    branches:
      - master
      - release/*
  pull_request:
    branches:
      - master
      - release/*

jobs:
  build_and_test:
    runs-on: ubuntu-latest

    strategy:
      matrix:
        aot: [false, true]

    env:
      DOTNET_NOLOGO: true
      DOTNET_CLI_TELEMETRY_OPTOUT: 1

    steps:
      - uses: actions/checkout@v4
        with:
          fetch-depth: 0

      - uses: actions/setup-dotnet@v4
        with:
          dotnet-version: '10.0.x'

      - name: Restore
        run: dotnet restore

      - name: Build
        run: dotnet build -c Release

      - name: Test (normal)
        if: matrix.aot == 'false'
        run: dotnet test -c Release --no-build

      - name: Publish AOT
        if: matrix.aot == 'true'
        run: dotnet publish test/NCalc.Tests/NCalc.Tests.csproj \
          -c Release \
          -r linux-x64 \
          -p:PublishAot=true \
          -p:StripSymbols=true \
          --self-contained true

      - name: Run AOT binary
        if: matrix.aot == 'true'
        run: ./test/NCalc.Tests/bin/Release/net10.0/linux-x64/publish/NCalc.Tests
```

This is because the test is running at the compiled artifact, so we need 2 artifacts.

gumbarros:
Just an idea, maybe for another PR, what if we use a source generator for repeated operations like MathHelper?

gumbarros:
And ready to merge?

vpenades:
> Just an idea, maybe for another PR, what if we use a source generator for repeated operations like MathHelper?

My understanding is that NCalc is useful for when the function to be parsed and evaluated is available at runtime, not at compile time, so I don't know if that would be of any use.

If what you're talking about is pre-generating common funcions, then I think a T4 template would be more suitable.

That or I completely misunderstood the point 😸 

> And ready to merge?

yes, I think so.

gumbarros:
> > Just an idea, maybe for another PR, what if we use a source generator for repeated operations like MathHelper?
> 
> My understanding is that NCalc is useful for when the function to be parsed and evaluated is available at runtime, not at compile time, so I don't know if that would be of any use.
> 
> If what you're talking about is pre-generating common funcions, then I think a T4 template would be more suitable.
> 
> That or I completely misunderstood the point 😸 
> 
> > And ready to merge?
> 
> yes, I think so.

Yeah, like pre-generation

Nice, I will merge soon

cuican6:
Any plans to merge this PR? I'm really interested in this feature – could you share the timeline? Happy to help with testing if needed. Thanks!

gumbarros:
I'm at vacations until next week, any maintainer feel free to merge @Bykiev

PR #524, opened 2025-12-08
[WIP] migrate to TUnit

Bykiev:
Migrating from xUnit to TUnit to support AOT in the future


gumbarros:
> @gumbarros, can you please help me with generation access token for codecov and adding it to secrets? It seems to be blocked for my country. Or maybe we can use ReportGenerator tool and host the report on GitHub pages?

Sure, sorry for the late response. I created the `CODECOV_TOKEN` secret in our organization.

PR #526, opened 2025-12-16
[WIP] Use Parlot source generator

Bykiev:


src/NCalc.Core/Parser/NCalcGrammar.cs
```diff
@@ -0,0 +1,292 @@
...
+
+            return OneOf(
+                Terms.Text("NOT", true)
+                .AndSkip(OneOf(Literals.WhiteSpace(), Not(AnyCharBefore(openParen)))),
+                Terms.Text("!"));
+        }
+
+        [GenerateParser]
+        public static Parser<string> AndParser() =>
+            OneOf(Terms.Text("AND", true), Terms.Text("&&"));
```
sebastienros: One thing to check in the generated source is that usages of OneOf/Or are converted to switch/cases. Not just this one. That will validate the decomposition of smaller generated parser is actually supported.

sebastienros: I think outer parsers might still generate and contain the same code ... Not a big deal but not sure having a generated version for each node makes sense. Can't be less than one, but less is better.

src/NCalc.Core/Parser/NCalcGrammar.cs
```diff
@@ -0,0 +1,292 @@
...
+
+namespace NCalc.Parser
+{
+    public static partial class NCalcGrammar
+    {
+        [GenerateParser]
+        public static Parser<long> HecOctBinNumberParser()
+        {
+            var hexNumber = Terms.Text("0x")
+                .SkipAnd(Terms.Pattern(c => "0123456789abcdefABCDEF".Contains(c)))
```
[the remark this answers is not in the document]
sebastienros: I need to revisit my assumptions ;) and maybe, like the compiler does, generate different logic based on the input.

sebastienros:
Update

I started a new branch because the extra Grammars file with all the externalized parsers was a pain to manage. So now it looks much like the original parser. I also updated Parlot to handle `Select` and `Switch` differently (index based, c.f. sourcegen PR in Parlot), because otherwise there would be no source gen for these as the parser is returned dynamically from the lambda (unknown at compile time).

It's working, the decimal separator and culture are fixed, but the benchmarks are still valid. I compared by commenting out the `[GenerateParser]` attribute, and keep Antlr as a baseline to be sure it's working. 30 tests are failing due to the culture/decimla separator convenvience change (27 only without sourcegen, which reflect the bug you mentioned earlier, will check later).

Withouth source gen:

```

| Method                   | Mean      | Error     | StdDev    | Gen0   | Gen1   | Allocated |
|------------------------- |----------:|----------:|----------:|-------:|-------:|----------:|
| SimpleParlotExpression   |  2.178 us | 0.0260 us | 0.0014 us | 0.1526 |      - |   1.27 KB |
| SimpleAntlrExpression    |  3.728 us | 0.1166 us | 0.0064 us | 1.6365 | 0.0687 |  13.38 KB |
| AdvancedParlotExpression |  5.398 us | 0.0707 us | 0.0039 us | 0.3662 |      - |      3 KB |
| AdvancedAntlrExpression  | 11.486 us | 0.4866 us | 0.0267 us | 4.6997 | 0.5035 |  38.45 KB |
```

With sourcegen:

```
| Method                   | Mean      | Error     | StdDev    | Gen0   | Gen1   | Allocated |
|------------------------- |----------:|----------:|----------:|-------:|-------:|----------:|
| SimpleParlotExpression   |  1.624 us | 0.0337 us | 0.0018 us | 0.1507 |      - |   1.23 KB |
| SimpleAntlrExpression    |  3.791 us | 0.1467 us | 0.0080 us | 1.6365 | 0.0687 |  13.38 KB |
| AdvancedParlotExpression |  4.314 us | 0.1638 us | 0.0090 us | 0.3662 |      - |   3.05 KB |
| AdvancedAntlrExpression  | 11.940 us | 1.1322 us | 0.0621 us | 4.6997 | 0.5035 |  38.45 KB |
```

This is 25% and 20% improvement in parsing time right now.

sebastienros:
Can be closed in favor or https://github.com/ncalc/ncalc/pull/533

PR #527, opened 2025-12-17
Disable parser loop detection

Bykiev:
As the parser is fully under our control infinite circle on parser level is impossible. Disabling this feature will improve the overall performance


Bykiev:
Thank you, have a nice holidays!

PR #536, opened 2026-01-09
Lambdas - allow to override if, in functions. Support for Expression…

Bykiev:
Lambdas - allow to override `if`, `in` functions. Support for` ExpressionOptions.IgnoreCaseAtBuiltInFunctions`.

Move `MathFunctionHelper` and `MathMethodInfo` into plugin.

This a a breaking change, I think it should be merged into v6.


gumbarros (approved):
Lgtm

PR #544, opened 2026-02-09
Added `Assert.Expression` extension to unit tests

gumbarros:


gumbarros:
There is lots of tests that can be simplified in another PR, but like they said in spanish, "Mucha chamba" 😄

PR #545, opened 2026-02-09
fix json serialisation example

joriszwart:


joriszwart:
Well, it compiles :-) See this issue:

- #552

PR #553, opened 2026-04-22
Migrate unit tests to TUnit

gumbarros:
This PR also fixes json serialization for `System.Text.Json`


gumbarros:
@Bykiev i did not used #524 because lots of conflicts

vpenades:
I see a lot of `#if NET #endif` blocks in the json serialization code....  is that because NCalc targets netstandard2.0 and net472?

In these cases, what I usually do is to conditionally include System.Text.Json as a package dependency only for those target frameworks.

gumbarros:
> I see a lot of `#if NET #endif` blocks in the json serialization code.... is that because NCalc targets netstandard2.0 and net472?
> 
> In these cases, what I usually do is to conditionally include System.Text.Json as a package dependency only for those target frameworks.

You are right, but the #ifdef were not needed anyway, look what I found 🤣

```shell
gumbarros@gumbarros:~/RiderProjects/ncalc/src/NCalc.Core$ dotnet nuget why System.Text.Json
Project 'NCalc.Core' has the following dependency graph(s) for 'System.Text.Json':

  [net462]                                                                                                                                                                          
  [net8.0]                                                                                                                                                                          
  [net9.0]                                                                                                                                                                          
  [netstandard2.0]                                                                                                                                                                  
  └── Microsoft.Extensions.Logging.Console (v10.0.2)                                                                                                                                
      └── System.Text.Json (v10.0.2)         
```

vpenades:
> > I see a lot of `#if NET #endif` blocks in the json serialization code.... is that because NCalc targets netstandard2.0 and net472?
> > In these cases, what I usually do is to conditionally include System.Text.Json as a package dependency only for those target frameworks.
> 
> You are right, but the #ifdef were not needed anyway, look what I found 🤣
> 
> ```shell
> gumbarros@gumbarros:~/RiderProjects/ncalc/src/NCalc.Core$ dotnet nuget why System.Text.Json
> Project 'NCalc.Core' has the following dependency graph(s) for 'System.Text.Json':
> 
>   [net462]                                                                                                                                                                          
>   [net8.0]                                                                                                                                                                          
>   [net9.0]                                                                                                                                                                          
>   [netstandard2.0]                                                                                                                                                                  
>   └── Microsoft.Extensions.Logging.Console (v10.0.2)                                                                                                                                
>       └── System.Text.Json (v10.0.2)         
> ```

yeah, System.Text.Json is quite ubiquous by now.

In any care, I would suggest to reference the package directly in the project even if it's a transitive dependency of Logging, there's two reasons for that:

1.- maybe MS.Logging removes Text.Json in the future (unlikely, but who knows, maybe they refactor it out to a support library)
2.- maybe yourself choose to no longer depend on MS.Logging

In fact, the recomended pattern is for libraries to use `Microsoft.Extensions.Logging.Abstractions` only, and let the main application to reference whichever logger sink it needs.  This is so because by enforcing a specific console or trace sinks into the dependencies, it may conflict with other libraries also doing the same.

gumbarros:
> > > I see a lot of `#if NET #endif` blocks in the json serialization code.... is that because NCalc targets netstandard2.0 and net472?
> > > In these cases, what I usually do is to conditionally include System.Text.Json as a package dependency only for those target frameworks.
> > 
> > 
> > You are right, but the #ifdef were not needed anyway, look what I found 🤣
> > ```shell
> > gumbarros@gumbarros:~/RiderProjects/ncalc/src/NCalc.Core$ dotnet nuget why System.Text.Json
> > Project 'NCalc.Core' has the following dependency graph(s) for 'System.Text.Json':
> > 
> >   [net462]                                                                                                                                                                          
> >   [net8.0]                                                                                                                                                                          
> >   [net9.0]                                                                                                                                                                          
> >   [netstandard2.0]                                                                                                                                                                  
> >   └── Microsoft.Extensions.Logging.Console (v10.0.2)                                                                                                                                
> >       └── System.Text.Json (v10.0.2)         
> > ```
> 
> yeah, System.Text.Json is quite ubiquous by now.
> 
> In any care, I would suggest to reference the package directly in the project even if it's a transitive dependency of Logging, there's two reasons for that:
> 
> 1.- maybe MS.Logging removes Text.Json in the future (unlikely, but who knows, maybe they refactor it out to a support library) 2.- maybe yourself choose to no longer depend on MS.Logging
> 
> In fact, the recomended pattern is for libraries to use `Microsoft.Extensions.Logging.Abstractions` only, and let the main application to reference whichever logger sink it needs. This is so because by enforcing a specific console or trace sinks into the dependencies, it may conflict with other libraries also doing the same.

Yeah, I completly agree with all points. I will work on using only abstractions for logging in another pr later, the library is using console for legacy reasons, there was a flag for console logging.

PR #554, opened 2026-04-23
Breaking change: Switch to `NullLoggerFactory` as default logger

gumbarros:


vpenades:
After removing console, I don't see any reference to system.text.json, was it added?

gumbarros:
> After removing console, I don't see any reference to system.text.json, was it added?

Yes, the other PR already added a explicit reference. .NET 6+ already contains system.text.json at itself.

Bykiev (approved):
Nice job, thank you!

PR #576, opened 2026-06-16
Feature/evaluate binary handler

george-domingos:
## Summary
Adds extensibility points for intercepting binary expression evaluation, allowing consumers to override or augment the result of binary operations (e.g. +, *, AND, OR) before NCalc's default evaluation logic runs.

## Changes
- Added `EvaluateBinaryHandler` and `EvaluateBinaryAsyncHandler` delegates, exposed via `Expression.EvaluateBinary` / `Expression.EvaluateBinaryAsync` events.
- Added `BinaryEventArgs`, which exposes the `BinaryExpression` being evaluated and lazy `LeftValue()/LeftValueAsync()` and `RightValue()/RightValueAsync()` accessors (each side is resolved lazy).
- `EvaluationVisitor.Visit(BinaryExpression, ...)` now raises `OnEvaluateBinary` before falling back to default evaluation; if a handler sets `args.Result`, that result is used and short-circuits the built-in logic.
- Added tests covering custom binary handlers (including overriding `+` and `*` semantics) and custom parameter types resolved through `EvaluateBinaryAsync`.

## Motivation
This mirrors the existing `EvaluateFunction`/`EvaluateParameter` extensibility pattern for binary operators, enabling use cases like custom operator semantics for domain-specific types, logging/auditing of intermediate evaluation steps, and short-circuiting expensive evaluations.

## Testing
Added `EventHandlersTests.ExpressionShouldEvaluateCustomBinaryHandler` and `ExpressionShouldEvaluateCustomBinaryCustomTypes`, both passing locally.

### DIFF
--- a/src/NCalc.Core/Expression.cs
+++ b/src/NCalc.Core/Expression.cs
@@ -44,6 +44,23 @@ public class Expression
     }
 
     /// <summary>
+    /// Event triggered to handle binary evaluation.
+    /// </summary>
+    public event EvaluateBinaryHandler EvaluateBinary
+    {
+        add => Context.EvaluateBinaryHandler += value;
+        remove => Context.EvaluateBinaryHandler -= value;
+    }
+    /// <summary>
+    /// Event triggered to handle binary evaluation.
+    /// </summary>
+    public event EvaluateBinaryAsyncHandler EvaluateBinaryAsync
+    {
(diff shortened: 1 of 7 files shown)


gumbarros:
Nice job. Thanks for the contribution. I was thinking if its also valid to add `EvaluateUnaryHandler` and `EvaluateTernaryHandler` but let someone need this use case and make a PR :)

PR #581, opened 2026-06-22
Add GetBuiltInFunctionNames method

george-domingos:
Added the GetBuiltInFunctionNames() method to publicly expose the list of supported built-in functions.
Motivation: Allows consumer applications to validate expressions preventively before execution or Power UI components for autocomplete (IntelliSense) or dynamic documentation.

Implementation Details:

Returns an immutable collection to prevent external modification.

Cached statically to avoid redundant memory allocations per call.


hao-zeng-tfs:
There’s no “sin” here. Is there a reason for that, or is it just a bug?

gumbarros:
> There’s no “sin” here. Is there a reason for that, or is it just a bug?

image

When you created the PR you missed "sin". My recommendation would be use the same method to declare the functions.

PR #582, opened 2026-06-22
Move 'if', 'in' functions to the top of Evaluate and EvaluateAsync in BultiIn

george-domingos:
Since logical and conditional operations are among the most frequently executed functions in expression evaluations, checking them first reduces string comparison overhead for the majority of execution paths.


gumbarros:
Good catch

PR #590, opened 2026-07-01
small fix for Ln support in NCalc.LambdaCompilation

gogolevsergey:
Small fix with missing function Ln in LambdaCompilation


gumbarros:
Thanks

PR #614, opened 2026-08-07
Fix: Honor StringComparer at `in` comparisons when NoStringTypeCoercion is set

### LINKED ISSUE #611: `in` stops honoring `StringComparer` when `NoStringTypeCoercion` is set (behavior change in 6.3.2)
**Describe the bug**
Between 6.3.1 and 6.3.2, `in` stopped applying `StringComparer` when both operands are strings and `NoStringTypeCoercion` is set. I couldn't tell from the changelog whether this was intended, so raising it rather than assuming.

**Example Expression**
```csharp
const ExpressionOptions options =
    ExpressionOptions.NoStringTypeCoercion | ExpressionOptions.CaseInsensitiveStringComparer;

new Expression("'md' in ('MD','DO')", options).Evaluate();  // 6.3.1: True | 6.3.2 - 7.0.2: False
new Expression("'md' == 'MD'",        options).Evaluate();  // True on all versions
```

Verified across 6.0.0, 6.1.0, 6.1.1, 6.2.0, 6.3.0, 6.3.1 (all `True`) and 6.3.2, 6.3.3, 6.4.0, 7.0.2 (all `False`). Mixed-type cases are unaffected — `1 in ('1')` correctly stays `false` with the flag set, on every version.

I can open a PR if the change was unintentional?

cokert:
When both operands are strings, comparing them involves no coercion to another type, so NoStringTypeCoercion should not change how they compare.
Since 6.3.2 the flag short-circuited string comparisons to ordinal equality, making `'md' in ('MD','DO')` false under a case-insensitive StringComparer while `'md' == 'MD'` stayed true.

Mixed-type operands are untouched, so disabling coercion still rejects `1 in ('1')` as before.

Fixes #611

### DIFF
--- a/src/NCalc.Core/Helpers/EvaluationHelper.cs
+++ b/src/NCalc.Core/Helpers/EvaluationHelper.cs
@@ -129,6 +129,9 @@ public static class EvaluationHelper
 
         if (noStringTypeCoercion)
         {
+            if (leftValue is string leftOnlyString && rightValue is string rightOnlyString)
+                return options.StringComparer.Equals(leftOnlyString, rightOnlyString);
+
             if (leftValue is string || rightValue is string)
                 return EqualityComparer<object?>.Default.Equals(leftValue, rightValue);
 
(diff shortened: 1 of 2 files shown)


gumbarros:
Perfect solution, thanks

gumbarros (approved):

PR #619, opened 2026-08-20
Added `DisallowSingleEquals` option

### LINKED ISSUE #618: Single equals '=' can be used for equality expression
## My feedback

I've noticed that NCalc supports a single equals `=` to be used in equality expressions. It appears to behave equivalently to double equals `==`.

```csharp
using NCalc;

Expression e = new("2 + 3 == 5");
Console.WriteLine(e.ToExpressionString() + " --> " + e.Evaluate());

// (2 + 3) = 5 --> True

e = new Expression("2 + 3 = 5");
Console.WriteLine(e.ToExpressionString() + " --> " + e.Evaluate());

// (2 + 3) = 5 --> True
```

This confuses me a bit. With most popular programming languages, only double equals `==` is used for equivalency check; single equals tends to be the assignment operator, by contrast. Because this is such a common concept, users of my NCalc-based app expect `"2 + 3 = 5"` to throw evaluation errors.

## The path forward

I would like to respect my users' expectation that single equals `=` cannot be used for equality check. Would it be possible to introduce a new [`MathOption`](https://github.com/ncalc/ncalc/blob/master/src/NCalc.Core/Helpers/MathOptions.cs) for this? I could use like:
```csharp
ExpressionConfiguration config = new()
{
    Evaluation = new ExpressionEvaluationOptions
    {
        Math = new MathOptions
        {
            AllowSingleEquals = false
        }
    }
};
```
The alternative, I suppose, is using a visitor and manually invalidating the expression if it involves a single equals `=`. Please let me know your thoughts.

gumbarros:
This closes #618

### DIFF
--- a/docs/articles/language/operators.md
+++ b/docs/articles/language/operators.md
@@ -105,6 +105,11 @@ These operators compare two values to check equality or inequality.
 * `=`, `==` : Equal to
 * `!=`, `<>` : Not equal to
 
+> **Note:** The single equals sign `=` is accepted as an equality operator for
+> convenience and legacy reasons, but you can require `==` by setting
+> `LogicalExpressionParserOptions.DisallowSingleEquals = true`. When disabled, expressions such
+> as `2 + 3 = 5` fail to parse, while `2 + 3 == 5` remains valid.
+
 **Examples:**
 ```csharp
 42 == 42            // true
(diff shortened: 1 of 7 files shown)


Patch4747:
I am happy with these changes, thank you. Implementing as a parser option makes sense to me.

gumbarros:
> I am happy with these changes, thank you. Implementing as a parser option makes sense to me.

Nice, soon I will release a new version

PR #622, opened 2026-09-07
Pin package versions

Bykiev:
For some reason I was unable to build the project in VS without FrozenSet changes and disabling IL2026;IL3050;RS1041 errors in source generator project.

This PR improves package versions management for different targets

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -1,32 +1,62 @@
 <Project>
-  <PropertyGroup>
-    <ManagePackageVersionsCentrally>true</ManagePackageVersionsCentrally>
-  </PropertyGroup>
-  <ItemGroup>
-    <PackageVersion Include="BenchmarkDotNet" Version="0.15.8" />
-    <PackageVersion Include="ExtendedNumerics.BigDecimal" Version="3003.2.0.161" />
-    <PackageVersion Include="FastExpressionCompiler" Version="5.4.1" />
-    <PackageVersion Include="Microsoft.AspNetCore.Components.WebAssembly" Version="10.0.10" />
-    <PackageVersion Include="Microsoft.AspNetCore.Components.WebAssembly.DevServer" Version="10.0.10" />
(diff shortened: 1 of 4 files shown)


Bykiev:
@sebastienros, can you please tell, is it possible to relax `System.Collections.Immutable` version in `Parlot` for old frameworks?

sebastienros:
"relax"? Meaning depending on a smaller patch version?
What problem do you see related to Parlot?

sebastienros:
Is that an issue because of parlot's sourcegen PR? Not sure based on the description

Bykiev:
> "relax"? Meaning depending on a smaller patch version? What problem do you see related to Parlot?

Yes, as you can see in this PR, for older targets I specified `Version="8.0.*"`, but `Parlot` depends on specific version  of `System.Collections.Immutable` package (>=10.0.8 in stable release). I see no issues, because .net 8 is finishing it's support circle, but when .net 11 will be available, you seems will update it too.

sebastienros:
Please correct me, it looks like you try to pin dependencies to match the target framework.
But for System.Collections.Immutable this is not necessary (99% certainty). It targets specific tfms without runtime specific dependencies (pulling higher runtime assets which could fail). That's why Parlot reference 10.0.latest.

Currently working on 2.0 to remove the ones that are not needed.
Would you want me to change these in 1.5.x still ?

Bykiev:
> Please correct me, it looks like you try to pin dependencies to match the target framework. But for System.Collections.Immutable this is not necessary (99% certainty). It targets specific tfms without runtime specific dependencies (pulling higher runtime assets which could fail). That's why Parlot reference 10.0.latest.
> 
> Currently working on 2.0 to remove the ones that are not needed. Would you want me to change these in 1.5.x still ?

You're right, then no need to change anything in v1.5

sebastienros:
https://www.nuget.org/packages/Parlot/1.5.9 has immutable collections dependency removed

